PR #11, opened 2013-12-10
Updates JSON output to include catalog information to fix #7

### LINKED ISSUE #7: Missing dcat:Catalog
There is no dcat:Catalog equivalent in the JSON representation of the datasets.  Perhaps this would be a useful place to store information about the total number available to also solve #6

rossjones:
Adding catalog metadata to the JSON output so that it more closely resembles https://github.com/okfn/ckanext-dcat/blob/master/examples/catalog.json

Also removes the 'q' element from the package_search as it is un-necessary (and *:* doesn't work for me).


rossjones:
Think we can safely close this now.

PR #28, opened 2015-04-28
DCAT RDF generation

amercader:
Placeholder issue for CKAN -> DCAT

### DIFF
--- a/ckanext/dcat/processors.py
+++ b/ckanext/dcat/processors.py
@@ -30,17 +30,11 @@ class RDFProfileException(Exception):
     pass
 
 
-class RDFParser(object):
-    '''
-    An RDF to CKAN parser based on rdflib
-
-    Supports different profiles which are the ones that will generate
-    CKAN dicts from the RDF graph.
-    '''
+class RDFProcessor(object):
 
     def __init__(self, profiles=None, compatibility_mode=False):
         '''
-        Creates a parser instance
+        Creates a parser or serializer instance
 
         You can optionally pass a list of profiles to be used.
 
@@ -94,6 +88,15 @@ class RDFParser(object):
 
         return profiles
 
+
+class RDFParser(RDFProcessor):
+    '''
+    An RDF to CKAN parser based on rdflib
+
+    Supports different profiles which are the ones that will generate
+    CKAN dicts from the RDF graph.
+    '''
+
     def _datasets(self):
         '''
         Generator that returns all DCAT datasets on the graph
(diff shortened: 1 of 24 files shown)


amercader:
Progress:

- [ ] Extend RDFParser and RDFProfile to generate graphs and serialize datasets / catalog
- [ ] Dataset endpoint
- [ ] Catalog endpoint
- [ ] Org endpoint
- [ ] Handle URIs: create suitable defaults, config option for base URI
- [ ] Better resource format support
- [ ] Spatial field on datasets
- [ ] JSON-LD support
- [ ] Docs
- [ ] Tests

If enough time
- [ ] Content negotiation

amercader:
@davidread @Hoedic This is still WIP but already functional. If you are interested feel free to have a play and let me know your thoughts

ckanext/dcat/processors.py
```diff
@@ -146,13 +152,123 @@ def datasets(self):
...
+
+    Supports different profiles which are the ones that will generate
+    the RDF graph.
+    '''
+
+    def graph_from_dataset(self, dataset_dict):
+        '''
+        Given a CKAN dataset dict, creates a graph using the loaded profiles
+
+        The class RDFLib graph (accessible via `parser.g`) will be updated by
```
davidread: serializer.g instead of parser.g?

amercader: you are right, forgot to update the docstrings on b2887d79b28bad634172b3f4ee657f65f793b3c2

ckanext/dcat/utils.py
```diff
@@ -0,0 +1,146 @@
...
+    Returns an URI for the dataset
+
+    This will be used to uniquely reference the dataset on the RDF
+    serializations.
+
+    The value will be the first found of:
+
+        1. The value of the `uri` field
+        2. The value of an extra with key `uri`
+        3. `catalog_uri()` + '/dataset/' + `id` field
```
davidread: I would have thought package.name was preferable to package.id. I realise that the name might change and "cool URIs don't change". But it's pretty rare. And Linked Data people really like readable URIs, and my feeling is that that is more important. I'd love to get an 'expert' opinion on this though.

davidread:
Aside from the minor comments I've left to ponder, this looks great. Loving the ability to customize the profile.

I'm not sure we would use yet though, until we can cache the results somewhere. I wonder if we shouldn't just stick the serializations in a table in the database. Or SOLR is pretty cache-like - we cache the package dict - but I can it's not ideal to include this in the package object too as it slows up the indexing of the catalogue further, and the dcat is only of secondary importance. Redis would mean it's kept in memory all the time which is unnecessary. So I'm favouring database.

ckanext/dcat/utils.py
amercader: @davidread I can see where you are coming from, I just feel that specially when dealing when harvested datasets coming from different sources you really want that to be immutable. +1 on getting the opinion of an expert if we can find one :)

Issue #31, opened 2015-06-05
process profiles in the order of declaration


pduchesne:
Currently profiles are taken in the order of their hosting extensions as sorted by the system (cf https://github.com/ckan/ckanext-dcat/blob/c8c7314c8f36b3c5c9ae136f36d48e1142744551/ckanext/dcat/parsers.py#L80#L80)
They should be ordered according to the config declaration ('ckanext.dcat.rdf.profiles').

amercader:
Fixed in #32

pduchesne:
For some merging reason, the above fix was applied on parsers.py, after the same file was renamed to processors.py in the revision history. As a result, the fix got lost and is missing in current master.

amercader:
Fixed, thanks @pduchesne

PR #32, opened 2015-06-05
[#31] sort profiles after the config declaration

pduchesne:
This solves the issue, although it may not be the most efficient way


amercader:
Thanks for the patch @pduchesne, it looks good, I just added a small change in e42d5d33 to make sure no two different extensions can have a profile with the same name, to avoid confusion.

PR #33, opened 2015-07-01
Allow rdf harvester after_download to run if there is no content

tryggvib:
The after_download hook is not run if the content is empty (some
sites return no content and status code 200 if a site doesn't
exist).

If the after_download hook for example does some validation on the
content and returns the validation errors via gather errors, then
returning before after_download is called means the validation will
not be run and it may look like everything validates just fine
which obviously it shouldn't.

The return based on no content should then happen after the
after_download hooks have been called for all plugin
implementations.


amercader:
Thanks for the patch @tryggvib

PR #35, opened 2015-08-17
RDF DCAT Harvester also supports non-XML formats.

DieterDePaepe:
I have adjusted the DCAT parsing using RDFLib to support other RDF formats, such as turtle or n3. Docs and tests have been updated as well.

### DIFF
--- a/ckanext/dcat/harvesters/base.py
+++ b/ckanext/dcat/harvesters/base.py
@@ -1,7 +1,6 @@
 import os
 import uuid
 import logging
-import json
 
 import requests
 import rdflib
@@ -13,7 +12,6 @@ from ckan import model
 
 from ckanext.harvest.harvesters import HarvesterBase
 from ckanext.harvest.model import HarvestObject, HarvestObjectExtra
-from ckanext.dcat.processors import RDFParserException, RDFParser
 
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


pietercolpaert:
+1! Would love to see this merged :)

amercader:
This is great @DieterDePaepe, and on first quick skim it looks great. I'll try to set some time aside this week to review it properly and merge it. Thanks!

/cc @tryggvib @brew

ckanext/dcat/harvesters/base.py
```diff
@@ -71,9 +78,10 @@ def _get_content(self, url, harvest_job, page=1):
 
                 if length >= self.MAX_FILE_SIZE:
                     self._save_gather_error('Remote file is too big.', harvest_job)
-                    return None
+                    return None, None
 
-            return content
+            _content_format = _content_format or r.headers.get('content-type').split(";", 1)[0]
```
amercader: This will fail if the remote server does not return a `Content-type` header

ckanext/dcat/processors.py
```diff
@@ -130,7 +130,7 @@ def parse(self, data, _format=None):
         # Apparently there is no single way of catching exceptions from all
         # rdflib parsers at once, so if you use a new one and the parsing
         # exceptions are not cached, add them here.
-        except (SyntaxError, xml.sax.SAXParseException), e:
+        except (SyntaxError, xml.sax.SAXParseException, rdflib.plugin.PluginException), e:
```
amercader: Does this catch if RDFLib does not support / understand the format provided?

amercader:
@DieterDePaepe This looks great. See some minor comments I added.

Also you can implement a `validate_config()` method on harvesters to make sure that the JSON object that users entered on the form is valid. [Here](https://github.com/ckan/ckanext-spatial/blob/master/ckanext/spatial/harvesters/base.py#L126) is an example. In our case we should check that the value of the `rdf_format` key is one of the keys or values of this dict in utils.py:

https://github.com/ckan/ckanext-dcat/blob/master/ckanext/dcat/utils.py#L11

Besides this, it's ready to go!

DieterDePaepe:
Good feedback, will work on it this or next week.
I think the PluginException handles unknown content-types, but I'll double check.

DieterDePaepe:
I think this covers everything. I've used the rdflib for validating the configuration rather than your suggestion, since a lot more formats are supported than the utils dict.

amercader:
Thanks for the patch @DieterDePaepe, all merged now

PR #52, opened 2015-12-01
Added organization support

pstarace:
Hi I added the logic to support organizations export
Paolo

### DIFF
--- a/ckanext/dcat/controllers.py
+++ b/ckanext/dcat/controllers.py
@@ -8,6 +8,7 @@ else:
     from ckan.lib.base import BaseController
 from ckan.controllers.package import PackageController
 from ckan.controllers.home import HomeController
+from ckan.controllers.organization import OrganizationController
 
 from ckanext.dcat.utils import CONTENT_TYPES, parse_accept_header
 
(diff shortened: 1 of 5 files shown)


ckanext/dcat/controllers.py
```diff
@@ -45,6 +46,26 @@ def read_catalog(self, _format=None):
...
+            _format = check_access_header()
+
+        if not _format:
+            return OrganizationController().read(_id)
+
+        toolkit.response.headers.update(
+            {'Content-type': CONTENT_TYPES[_format]})
+
+        try:
+            fq = ' organization:"'+_id+'"'
```
amercader: Don't add the `fq` param here, just pass the `id` to the action (see comment below)

ckanext/dcat/logic.py
```diff
@@ -30,6 +29,21 @@ def dcat_dataset_show(context, data_dict):
 
     return output
 
+def dcat_organization_show(context, data_dict):
+
+    toolkit.check_access('dcat_organization_show', context, data_dict)
+
+    query = _search_ckan_datasets(context, data_dict)
```
amercader: Add the `fq` parameter here, that makes the API for the action cleaner (you just pass the `id` of the organization).
Also use this as the value for `fq`, to allow using the name (what's on the URL) and the internal id:

```
'organization:{id} OR owner_org:{id}'.format(id=_id)
```

amercader:
Thanks @pstarace this looks really good. Just see my small comments, and also could you add a couple of tests similar to this one?

https://github.com/ckan/ckanext-dcat/blob/master/ckanext/dcat/tests/test_controllers.py#L190

amercader:
And if you merge master you will get rid of these test failures as I fixed them now

Issue #57, opened 2016-01-30
JSON-LD not advertised in HTML


neothemachine:
In all the templates at https://github.com/ckan/ckanext-dcat/tree/master/ckanext/dcat/templates there is no link to the JSON-LD resources. Is that just an oversight? If so, I'm happy to create a pull request to add that.

I also noticed in [home/index.html](https://github.com/ckan/ckanext-dcat/blob/master/ckanext/dcat/templates/home/index.html) there are two `application/rdf+xml` links, one going to `.rdf`, the other to `.xml` but their content is identical.  In the [package/read_base.html](https://github.com/ckan/ckanext-dcat/blob/master/ckanext/dcat/templates/package/read_base.html) template this is not the case. But then again in the [package/search.html](https://github.com/ckan/ckanext-dcat/blob/master/ckanext/dcat/templates/package/search.html) template there are two xml ones. This should be fixed as well, right?

amercader:
@neothemachine these are indeed oversights, if you can send a PR adding the link to JSON-LD and removing the duplicated `.rdf` one I'll be happy to merge it

amercader:
Fixed in #58

PR #58, opened 2016-02-01
fix <link>s in templates

neothemachine:
See #57


amercader:
Thanks for the patch @neothemachine

PR #59, opened 2016-02-25
Use https URL for geomet dependency

dlax:
https:// URL has more chance to be reachable that git:// one, e.g., when
behind a firewall.


amercader:
Thanks for the patch @dlax

PR #65, opened 2016-06-22
Add a gather error instead of log.error

metaodi:
A missing guid is a clear indicator of a source that is broken.
These errors should be visible and not buried in an error log.


amercader:
Sounds good but there is a syntax error

amercader:
```
from ckanext.dcat.harvesters.rdf import DCATRDFHarvester

  File "/home/travis/build/ckan/ckanext-dcat/ckanext/dcat/harvesters/rdf.py", line 205

    harvest_job)

    ^

IndentationError: unexpected indent
```

metaodi:
@amercader oh no! Just fixed it :wink:

PR #69, opened 2016-09-14
Use ckan.site_url if available, request.host_url as fallback

### LINKED ISSUE #68: Pagination uses toolkit.request.path_url instead of ckan.site_url
We're using the catalog output of ckanext-dcat for a CKAN instance and just realized, that the [pagination with HYDRA uses `toolkit.request.path_url`](https://github.com/ckan/ckanext-dcat/blob/54f6366a5c6368e70cc136b32fe7b2006f30ee27/ckanext/dcat/logic.py#L152) to determine the URL of the page. Because of our server/network setup, this now returns an internal URL, which is not accessible from the internet (the server is behind a proxy).

The obvious solution would be to use the CKAN option `ckan.site_url` instead to make sure, the URL is always the external one. Should the code be change to use this option? Or is there a specific use case where `toolkit.request.path_url` is the better option?

Alternatively we could introduce a new ckanext-dcat specific option to control wheter to use `site_url` or not. Something like `ckanext.dcat.always_use_site_url` with a boolean value.

metaodi:
Fixes #68

### DIFF
--- a/ckanext/dcat/logic.py
+++ b/ckanext/dcat/logic.py
@@ -147,8 +147,8 @@ def _pagination_info(query, data_dict):
         base_url = config.get('ckan.site_url', '').strip('/')
         if not base_url:
             base_url = toolkit.request.host_url
-        base_url = '%s%s%s' % (
-            base_url, toolkit.request.path_info, toolkit.request.script_name)
+        base_url = '%s%s' % (
+            base_url, toolkit.request.path)
 
         params = [p for p in toolkit.request.params.iteritems()
                   if p[0] != 'page']
--- a/ckanext/dcat/logic.py
+++ b/ckanext/dcat/logic.py
@@ -144,18 +144,24 @@ def _pagination_info(query, data_dict):
 
     def _page_url(page):
 
+        base_url = config.get('ckan.site_url', '').strip('/')
+        if not base_url:
+            base_url = toolkit.request.host_url
+        base_url = '%s/%s%s' % (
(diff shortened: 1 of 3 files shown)


metaodi:
@amercader can you take a look at this? The PR basically uses `ckan.site_url` if it is available, to make sure the returned URL of the pagination is the external one. The previously used [`path_url` is a combination of the request host, the path and the script name](http://docs.pylonsproject.org/projects/pyramid/en/latest/api/request.html#pyramid.request.Request.path_url), that's why I include those values again after setting the correct domain.

ckanext/dcat/logic.py
```diff
@@ -144,18 +144,24 @@ def _pagination_info(query, data_dict):
 
     def _page_url(page):
 
+        base_url = config.get('ckan.site_url', '').strip('/')
+        if not base_url:
+            base_url = toolkit.request.host_url
+        base_url = '%s%s%s' % (
+            base_url, toolkit.request.path_info, toolkit.request.script_name)
```
amercader: `request.host_url` works on both, but `request.path_info` and `request.script_name` are not present on the Flask request object. Both Flask's and Pylons's have `request.path` though, which is the combination of both (eg `/feed/catalog.xml` for http://example.com/feed/catalog.xml`). Let's use that and future-proof this! :rocket:

metaodi: @amercader I finally got around to fix this use `path`.

amercader:
Thanks @metaodi!

PR #71, opened 2016-09-29
Vg/patch missing exports

vgammieri:
I found two missing fields on the output, so I added it to the default profile


amercader:
Thanks a lot @vgammieri!

PR #76, opened 2016-12-02
Add MANIFEST.in to ensure i18n files are include in package

insertjokehere:
Without this, if you install the package not in 'editable' (ie,`setup.py install` instead of `setup.py develop` or `pip install -e`) mode, the i18n files are not installed, and the extension cannot be enabled


amercader:
Thanks for the patch @insertjokehere

PR #86, opened 2017-01-31
Ignore certain datasets when harvesting

### LINKED ISSUE #85: Ignore certain datasets when harvesting
I want to ignore some datasets when harvesting and I was thinking of using the `before_create` and `before_update` methods of the `IDCATRDFHarvester` interface.

Of course I could just empty the dataset dict and then the `package_create` and `package_update` would fail with an error. But I'd much rather [enhance the code](https://github.com/ckan/ckanext-dcat/blob/4e58a5158621c9c6d326c0fa6fbf604ea56d4851/ckanext/dcat/harvesters/rdf.py#L354) with something like this to avoid the error:

```python
if dataset:
   p.toolkit.get_action('package_create')(context, dataset)
else:
   log.info('There is no dataset to create, continue')
   return True
```

Or maybe add a boolean return value to these two methods and check this. What do you think?

metaodi:
Fixes #85

### DIFF
--- a/ckanext/dcat/harvesters/rdf.py
+++ b/ckanext/dcat/harvesters/rdf.py
@@ -309,7 +309,11 @@ class DCATRDFHarvester(DCATHarvester):
                 harvester.before_update(harvest_object, dataset, harvester_tmp_dict)
 
             try:
-                p.toolkit.get_action('package_update')(context, dataset)
+                if dataset:
+                    p.toolkit.get_action('package_update')(context, dataset)
+                else:
+                    log.info('Ignoring dataset %s' % existing_dataset['name'])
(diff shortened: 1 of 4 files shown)


amercader:
This approach looks good @metaodi :+1: 

You need to `p.load()` the plugin on the test to make it active: https://github.com/ckan/ckanext-dcat/blob/f8360181f72acf88093a0e1651b281717fe6faee/ckanext/dcat/tests/test_harvester.py#L778

metaodi:
@amercader thanks for the hint. After some back and forth I could finally get the tests to work :wink:

amercader:
Thanks @metaodi, great stuff

PR #88, opened 2017-02-20
Retrieve license information from distribution

### LINKED ISSUE #42: DCAT and CKAN disagree on license location
DCAT defines the license element as a Distribution property (i.e. part of a Resource in CKAN terms), but CKAN defines the license at the Dataset level.
This is a structural mismatch that's probably beyond the scope of this component, but it is still an issue in its usage. Can ckanext-dcat offer a pragmatic, workaround solution, like supporting a dct:license field on dcat:Dataset, or take the first license of a resource as the dataset license ?

dlax:
Closes #42

Following a recent [thread](https://lists.okfn.org/pipermail/ckan-dev/2017-February/010794.html) on the mailing list.
We have been using this patch for a couple of times; I finally got to finish the patch and submit.

### DIFF
--- a/ckanext/dcat/profiles.py
+++ b/ckanext/dcat/profiles.py
@@ -11,6 +11,7 @@ from rdflib.namespace import Namespace, RDF, XSD, SKOS, RDFS
 
 from geomet import wkt, InvalidGeoJSONException
 
+from ckan.model.license import LicenseRegister
 from ckan.plugins import toolkit
 
 from ckanext.dcat.utils import resource_uri, publisher_uri_from_dataset_dict
@@ -609,6 +610,13 @@ class EuropeanDCATAPProfile(RDFProfile):
             if value:
                 dataset_dict['version'] = value
 
+        # Mappings for license retrieval (from distribution to dataset)
+        license_uri2id = {}
+        license_title2id = {}
+        for license_id, license in LicenseRegister().items():
+            license_uri2id[license.url] = license_id
+            license_title2id[license.title] = license_id
+
         # Tags
         keywords = self._object_value_list(dataset_ref, DCAT.keyword) or []
         # Split keywords with commas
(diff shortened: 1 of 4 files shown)


ckanext/dcat/profiles.py
```diff
@@ -769,6 +777,19 @@ def parse_dataset(self, dataset_dict, dataset_ref):
...
+                license = self._object(distribution, DCT.license)
+                if license:
+                    # Try to find a matching license comparing URIs, then titles
+                    license_id = license_uri2id.get(license.toPython())
+                    if license_id is None:
+                        license_id = license_title2id.get(
+                            self._object_value(license, DCT.title))
+                    if license_id is not None:
+                        dataset_dict['license_id'] = license_id
+
```
metaodi: I would propose to put this whole block in a method called `_license()` to make it very easy to override this behavior.

dlax: Done in de220a425c2fb08edf6dcd2b4db3bf7ea5ba5336

ckanext/dcat/profiles.py
```diff
@@ -307,6 +308,35 @@ def _spatial(self, subject, predicate):
...
+        found in CKAN license registry. If no distribution's license matches,
+        None is returned.
+
+        The first distribution with a license found in the registry is used so
+        that if distributions have different licenses we'll only get the first
+        one.
+        '''
+        license_uri2id = {}
+        license_title2id = {}
+        for license_id, license in LicenseRegister().items():
```
amercader: This only needs to be run once right? There's no need to run the same loop on every dataset (and there could be potentially loads of them when harvesting). So you could do something like:

```python 
if not self._license_uri2id:
    for license_id, license in LicenseRegister().items():
         self._license_uri2id[license.url] = license_id
         # ...
```

dlax:
Adrià Mercader a écrit :
> Thanks @dlax <https://github.com/dlax>, see just one minor comment. Also
> this should definitely be documented in the README. I suggest adding a
> note on the mappings table
> <https://github.com/ckan/ckanext-dcat#rdf-dcat-to-ckan-dataset-mapping>
> ("See note about dataset license") and expand on what's the default
> behaviour and that can people can customize it overriding the |_license|
> method on their own profile.

Added two commits addressing these points.

amercader:
Thanks @dlax, great stuff

PR #90, opened 2017-03-28
Removing IPackageController code

thriuin:
Removing code that does not appear to be used and is interfering with rendering the read package page on our web site at open.canada.ca

### DIFF
--- a/ckanext/dcat/logic.py
+++ b/ckanext/dcat/logic.py
@@ -1,5 +1,6 @@
 from __future__ import division
 import math
+import json
 
 from pylons import config
 from dateutil.parser import parse as dateutil_parse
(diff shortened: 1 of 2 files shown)


wardi:
interesting that it only fails on more recent ckan versions. @amercader the `after_show` code is interfering with our custom schema (we have a language field at the resource level) do you know why it's implemented this way and not with a helper and an overridden resource template?

amercader:
@wardi as this is a "generic" extension I didn't want to override the templates as users might have already done so, but if necessary I'm willing to consider it.

In theory `after_show` hooks are called in succession so if your extension if loaded after `dcat` can you not modify the data_dict according to your needs?

wardi:
@amercader this PR was supposed to only include 5ababd0 we can open a new one.

thriuin:
Ah, my bad, I merged in those last 2 commits yesterday - they shouldn't be there.

PR #94, opened 2017-08-17
Map existing resources based on their URI

### LINKED ISSUE #91: resources are re-created on reharvest
It seems that upon reharvest of a DCAT source, resources of dataset are deleted and recreated.

This can be seen in the following excerpt from "fetch consumer" process logs:
```
2017-05-03 07:52:43,487 INFO  [ckanext.harvest.queue] Received harvest object id: 1e42effb-7dbb-492a-b8c8-e0681be0c287
2017-05-03 07:52:43,510 DEBUG [ckanext.dcat.harvesters.rdf] In DCATRDFHarvester import_stage
2017-05-03 07:52:43,759 DEBUG [ckanext.datapusher.plugin] Submitting resource e906f816-14af-40c3-99c8-42f10e8ec2f1 to DataPusher
2017-05-03 07:52:43,879 DEBUG [ckanext.datapusher.plugin] Submitting resource 8ba84cb8-dcfa-4404-9ea5-529e39e1eebc to DataPusher
2017-05-03 07:52:43,995 DEBUG [ckanext.datapusher.plugin] Submitting resource c028af71-86f8-4735-86f6-164b96ffe28b to DataPusher
2017-05-03 07:52:44,214 DEBUG [ckanext.archiver.plugin] Notified of package event: liste-des-marches-publics-conclus-par-le-departement-de-la-gironde changed
2017-05-03 07:52:44,234 DEBUG [ckanext.archiver.plugin] Comparing with revision: 2017-05-03 07:27:29.780708 94a67ddc-b973-41e5-ac88-8bc9fce4c116
2017-05-03 07:52:44,306 DEBUG [ckanext.archiver.plugin] Deleted resources - will archive. res_ids=set([u'2bd92e64-a138-46c3-a17f-962d7468ad5f', u'c3711e80-1514-4409-944f-27999512d1a6', u'bdb4934c-ade4-48c2-a97c-501a8d7f7fdf'])
2017-05-03 07:52:44,306 DEBUG [ckanext.archiver.plugin] Creating archiver task: liste-des-marches-publics-conclus-par-le-departement-de-la-gironde
2017-05-03 07:52:44,310 DEBUG [ckanext.archiver.lib] Archival of package put into celery queue priority: liste-des-marches-publics-conclus-par-le-departement-de-la-gironde
2017-05-03 07:52:44,333 INFO  [ckanext.dcat.harvesters.rdf] Updated dataset liste-des-marches-publics-conclus-par-le-departement-de-la-gironde
2017-05-03 07:52:44,356 INFO  [ckanext.harvest.queue] Received harvest object id: 09cb47c9-f93f-46aa-88a1-0ca2b62760ac
2017-05-03 07:52:44,380 DEBUG [ckanext.dcat.harvesters.rdf] In DCATRDFHarvester import_stage
```

As can be seen, my setup includes datapusher/datastore plugins.

I noticed that, on every reharvest, prior resources are not anymore available even though nothing changed in the source in the meantime; in particular, their uuid changed so that one gets a 404 on old resources URL.

It's also worrying as it seems that resources are not cleared from the datastore as the datapusher actually accumulated them. Upon every reharvest, in datapusher's process log, I get a message like:
```
Successfully pushed 62220 entries to "937a73d6-ba5b-46db-934e-944ca65eb2b2".
```

metaodi:
Fixes #91

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ install:
     - bash bin/travis-build.bash
 services:
     - redis-server
+    - postgresql
 script: sh bin/travis-run.sh
 before_install:
     - pip install codecov
--- a/bin/travis-build.bash
+++ b/bin/travis-build.bash
@@ -5,7 +5,7 @@ echo "This is travis-build.bash..."
 
 echo "Installing the packages that CKAN requires..."
 sudo apt-get update -qq
-sudo apt-get install postgresql-9.1 solr-jetty
+sudo apt-get install solr-jetty
 
 echo "Installing CKAN and its Python dependencies..."
 git clone https://github.com/ckan/ckan
--- a/bin/travis-build.bash
+++ b/bin/travis-build.bash
@@ -29,8 +29,8 @@ python setup.py develop
 # Pin this as newer versions installed by RDFLib give setuptools troubles
 pip install "html5lib==0.9999999"
 
-pip install -r requirements.txt --allow-all-external
-pip install -r dev-requirements.txt --allow-all-external
+pip install -r requirements.txt
+pip install -r dev-requirements.txt
 cd -
 
 echo "Setting up Solr..."
@@ -51,13 +51,13 @@ echo "Installing ckanext-harvest and its requirements..."
 git clone https://github.com/ckan/ckanext-harvest
 cd ckanext-harvest
 python setup.py develop
-pip install -r pip-requirements.txt --allow-all-external
+pip install -r pip-requirements.txt
 paster harvester initdb -c ../ckan/test-core.ini
 cd -
 
 echo "Installing ckanext-dcat and its requirements..."
-pip install -r requirements.txt --allow-all-external
-pip install -r dev-requirements.txt --allow-all-external
+pip install -r requirements.txt
+pip install -r dev-requirements.txt
 python setup.py develop
 
 
--- a/ckanext/dcat/harvesters/rdf.py
+++ b/ckanext/dcat/harvesters/rdf.py
@@ -311,6 +311,14 @@ class DCATRDFHarvester(DCATHarvester):
 
                 harvester_tmp_dict = {}
 
+                # check if resources already exist based on their URI
+                existing_resources =  existing_dataset.get('resources')
+                resource_mapping = {r.get('uri'): r.get('id') for r in existing_resources if r.get('uri')}
+                for resource in dataset.get('resources'):
+                    res_uri = resource.get('uri')
+                    if res_uri and res_uri in resource_mapping:
+                        resource['id'] = resource_mapping[res_uri]
+
                 for harvester in p.PluginImplementations(IDCATRDFHarvester):
                     harvester.before_update(harvest_object, dataset, harvester_tmp_dict)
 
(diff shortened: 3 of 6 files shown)


metaodi:
This PR also fixes the controller tests, which now require a test_request_context (see https://github.com/ckan/ckan/pull/3228). This is unrelated, but the build fails otherwise.

metaodi:
@amercader When writing tests I noticed something strange: if the URI of a resource is empty, it is saved with the string `"None"` is CKAN. I guess it's because it gets [explicitly assigned the value `None`](https://github.com/ckan/ckanext-dcat/blob/ef66a54a7fddd54e851c8559111559e2718a1bcd/ckanext/dcat/profiles.py#L807). I guess later on this is being serialized using `str()` or something similar, which then creates the string `"None"`.

In this PR, I now changed the value from `None` to an empty string. I think this is actually a very good solution, as you end up with an empty string in CKAN and all the conditionals (`if value` etc.) still work as the empty string is falsy in Python. WDYT?

The other solution would be to change to serialization, and make sure `None` values are saved as `null` or an empty string.

It probably would make sense to implement this consequently, as we have now empty strings and `None` values.

metaodi:
See https://github.com/ckan/ckan/issues/3780 for backwards-compatible `test_request_context`

metaodi:
ping @amercader

metaodi:
So I just removed my changed for the `test_request_context` since https://github.com/ckan/ckan/pull/3782 has been merged.

metaodi:
@amercader do you have time to review this PR?

amercader:
sorry @metaodi I've been a bit swamped recently. This looks great. What's the reason behind 31042e69?

metaodi:
@amercader [see my comment above](https://github.com/ckan/ckanext-dcat/pull/94#issuecomment-324267731), let me know if you need more information.

amercader:
Sorry @metaodi, should have read first. Your approach sounds good, we should definitely consolidate the behaviour in the serialization code but that can be done at another time.

Shall we change [this](https://github.com/ckan/ckanext-dcat/blob/31042e692458154544def0f3abf7a7ec8f96f91e/ckanext/dcat/profiles.py#L225) and [this](https://github.com/ckan/ckanext-dcat/blob/31042e692458154544def0f3abf7a7ec8f96f91e/ckanext/dcat/profiles.py#L252) as well for consistency?

metaodi:
@amercader okay I updated the other URIs as well, and to be consistent I updated `_object_value` to return an empty string if the value is not found, e.g. `_contact_details()` should return an empty string for all attributes not only `uri`, do you agree?

amercader:
Sounds good @metaodi, many thanks for this!

camfindlay:
Just checking - does this apply to the DCAT JSON Harvester or will that need to implemented still?

amercader:
@camfindlay the DCAT JSON harvester hasn't been updated in a long time. Current efforts are focused on the RDF based functionality.

camfindlay:
thanks @amercader - we're making use of DCAT JSON so take it if we'd like this same improvement we'll need to raise a PR and add a similar check to avoid harvesting when resources haven't changed. ^ CC/ @dythya

PR #95, opened 2017-08-29
Improve the error handling in the gather and import stage

metaodi:
- Raise exceptions when `parse_dataset` in profiles fails
- Do not update content hash if download errored
- Call `_save_gather_error` if `datasets()` generators fails

### DIFF
--- a/bin/travis-build.bash
+++ b/bin/travis-build.bash
@@ -14,6 +14,14 @@ if [ $CKANVERSION != 'master' ]
 then
     git checkout release-v$CKANVERSION-latest
 fi
+
+# install the recommended version of setuptools
+if [ -f requirement-setuptools.txt ]
+then
+    echo "Updating setuptools..."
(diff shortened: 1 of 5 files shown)


metaodi:
I added the traceback as we most likely have an error in a custom profile, so this makes the it easier pinpoint the source of the error (in my case I was hunting a `KeyError` somewhere in my custom profile)

metaodi:
I added two test cases to test the new error handling

1. The user-contributed profiles are used (`parse_dataset`), hence we don't have control over the code. If an exception is raised, the harvester should fail gracefully incl. a nice error msg in the log
2. The `IDCATRDFHarvester` provides extension points to the harvester. As we don't have control over the code behind, the same as above applies: if an exception is raised, the harvester should fail gracefully (possibly only for one harvest_object) and log an error msg.

amercader:
Thanks @metaodi!

PR #97, opened 2017-10-11
README: fix parameter name modified_date

seitenbau-govdata:
Change all occurrences of 'modified_date' to 'modified_since'.


amercader:
Thanks @seitenbau-govdata, good catch!

PR #99, opened 2017-10-24
Add `update_session` to interface

### LINKED ISSUE #98: Interface for `requests` sessions
For a client, I need to perform HTTP requests using a client certificate. This is something that the [`requests` library supports](http://docs.python-requests.org/en/master/user/advanced/#client-side-certificates), so I was thinking how we could implement that.

I propose to start using [`requests` Session objects](http://docs.python-requests.org/en/master/user/advanced/#session-objects) to make the remote HTTP calls. Then we could provide a new method on the `IDCATRDFHarvester` interface called e.g. `update_session()` which would receive an "empty" session object (or the session object changed by the last implementation of `update_session()` for subsequent calls).

This would enable us to write extensions that implement this interface, where each could update something in the session object, which then gets used to make the remote calls.

metaodi:
Fixes #98

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,8 +230,12 @@ The harvester will look at the `content-type` HTTP header field to determine the
 ### Extending the RDF harvester
 
 The DCAT RDF harvester has extension points that allow to modify its behaviour from other extensions. These can be used by extensions implementing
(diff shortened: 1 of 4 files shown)


amercader:
This looks great @metaodi, it's a really nice addition. It could be worth adding to other harvesters, so we know where to look of this is ever needed.

Issue #96, opened 2017-10-09
Source Catalog info in transitive harvesting


etj:
In transitive harvesting, source Catalog info is lost.

E.g.: Let's say node A and node B are two nodes in which datasets are created.
B harvests from A, so B contains datasets created locally, plus datasets created by A.

We then have a node C that harvests B. How can C find out which datasets belong to (i.e. were originally created on) B, and which datasets belong to A?
If we rely on the Catalog node only, it will seem that ALL datasets belong to catalog B, which isn't the case. So we should associate a new node to the dataset, that tells which is its original source catalog.

A solution is about using the `dct:hasPart` property, which expresses in pure DCAT semantic this requirement.

Let's say 
- Catalog C1 has Datasets D1a D1b
- Catalog C2 has Datasets D2a D2b

C1 exposes a catalog.rdf as:
- Catalog  C1 
  - Dataset D1a     
  - Dataset D1b

C2 harvests from C1.
The related RDF could be expressed as:
- Catalog C2 (URI, title, description, homepage)   
  - Dataset D2a        
  - Dataset D2b     
  - `dct:hasPart`
     - Catalog C1 (URI, title, description, homepage)        
       -  Dataset D1a               
       -  Dataset D1b

(Original discussion was in geosolutions-it/ckanext-dcatapit#52, but ckanext-dcat is better suited for implementing this improvement.)

PR #100, opened 2017-10-27
handle source catalog with dct:hasPart

cezio:
this implements https://github.com/ckan/ckanext-dcat/issues/96

### DIFF
--- a/ckanext/dcat/plugins.py
+++ b/ckanext/dcat/plugins.py
@@ -141,15 +141,3 @@ class DCATJSONInterface(p.SingletonPlugin):
         return {
             'dcat_datasets_list': dcat_auth,
         }
-
-
-class DCATFacetsPlugin(p.SingletonPlugin, DefaultTranslation):
-    
-    p.implements(p.IFacets, inherit=True)
-    if p.toolkit.check_ckan_version(min_version='2.5.0'):
-        p.implements(p.ITranslation, inherit=True)
-
-    # IFacets
-    def dataset_facets(self, facets_dict, package_type):
-        facets_dict['source_title'] = p.toolkit._("Sources")
-        return facets_dict
--- a/ckanext/dcat/plugins.py
+++ b/ckanext/dcat/plugins.py
@@ -26,16 +26,10 @@ class DCATPlugin(p.SingletonPlugin, DefaultTranslation):
     p.implements(p.IConfigurer, inherit=True)
     p.implements(p.IRoutes, inherit=True)
     p.implements(p.IActions, inherit=True)
-    p.implements(p.IFacets, inherit=True)
     p.implements(p.IAuthFunctions, inherit=True)
     p.implements(p.IPackageController, inherit=True)
     if p.toolkit.check_ckan_version(min_version='2.5.0'):
         p.implements(p.ITranslation, inherit=True)
-
-    # IFacets
-    def dataset_facets(self, facets_dict, package_type):
-        facets_dict['source_title'] = p.toolkit._("Sources")
-        return facets_dict
 
     # IConfigurer
     def update_config(self, config):
@@ -147,3 +141,15 @@ class DCATJSONInterface(p.SingletonPlugin):
         return {
             'dcat_datasets_list': dcat_auth,
         }
+
+
+class DCATFacetsPlugin(p.SingletonPlugin, DefaultTranslation):
+    
+    p.implements(p.IFacets, inherit=True)
+    if p.toolkit.check_ckan_version(min_version='2.5.0'):
+        p.implements(p.ITranslation, inherit=True)
+
+    # IFacets
+    def dataset_facets(self, facets_dict, package_type):
+        facets_dict['source_title'] = p.toolkit._("Sources")
+        return facets_dict
--- a/ckanext/dcat/plugins.py
+++ b/ckanext/dcat/plugins.py
@@ -26,10 +26,16 @@ class DCATPlugin(p.SingletonPlugin, DefaultTranslation):
     p.implements(p.IConfigurer, inherit=True)
     p.implements(p.IRoutes, inherit=True)
     p.implements(p.IActions, inherit=True)
+    p.implements(p.IFacets, inherit=True)
     p.implements(p.IAuthFunctions, inherit=True)
     p.implements(p.IPackageController, inherit=True)
     if p.toolkit.check_ckan_version(min_version='2.5.0'):
         p.implements(p.ITranslation, inherit=True)
+
+    # IFacets
+    def dataset_facets(self, facets_dict, package_type):
+        facets_dict['source_title'] = p.toolkit._("Sources")
+        return facets_dict
 
     # IConfigurer
     def update_config(self, config):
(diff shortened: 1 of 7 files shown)


ckanext/dcat/plugins.py
```diff
@@ -26,11 +26,17 @@ class DCATPlugin(p.SingletonPlugin, DefaultTranslation):
     p.implements(p.IConfigurer, inherit=True)
     p.implements(p.IRoutes, inherit=True)
     p.implements(p.IActions, inherit=True)
+    p.implements(p.IFacets, inherit=True)
     p.implements(p.IAuthFunctions, inherit=True)
     p.implements(p.IPackageController, inherit=True)
     if p.toolkit.check_ckan_version(min_version='2.5.0'):
         p.implements(p.ITranslation, inherit=True)
 
+    # IFacets
```
amercader: I would prefer if this was handled by each CKAN individual instance if they want to, rather than by ckanxet-dcat

ckanext/dcat/processors.py
```diff
@@ -5,6 +5,7 @@
 from pkg_resources import iter_entry_points
 
 from pylons import config
+from paste.deploy.converters import asbool
```
amercader: Please use the `asbool` function from the plugins toolkit.

amercader:
@cezio @etj I added some minor comments but, I need some time to properly review this, will try to do it next week.

As with all DCAT issues available documentation is very sparse and all suggestions I found seem to point that `dct:hasPart` is a property of catalog (ie Catalog -> catalog, not dataset -> catalog) so I wonder if there is another way to describe this relation better. In principle I'm happy to store provenance information in the datasets extras though.

@metaodi do you guys have any experience with this?

etj:
@amercader with this implementation, in DCAT output we'll have Catalog -> hasPart -> Catalog -> dataset indeed, as per DCAT documentation.
We're storing the catalog info in datasets' extras in order not to modify the dataset model.

ckanext/dcat/plugins.py
```diff
@@ -147,3 +141,15 @@ def get_auth_functions(self):
         return {
             'dcat_datasets_list': dcat_auth,
         }
+
+
+class DCATFacetsPlugin(p.SingletonPlugin, DefaultTranslation):
```
etj: As [requested by Adrian](https://github.com/ckan/ckanext-dcat/pull/100#discussion_r147440145), this faceting should not belong here. Please move it to the IT profile.

etj:
I created a new PR https://github.com/ckan/ckanext-dcat/pull/101 containing the squashed commits of this PR, plus some more cleanup and docs.

ckanext/dcat/profiles.py
```diff
@@ -570,6 +573,36 @@ def _last_catalog_modification(self):
...
+        Catalog referenced in dct:hasPart is returned, 
+        if dataset is linked there, otherwise main catalog 
+        will be returned.
+
+        This will not be used if ckanext.dcat.expose_subcatalogs
+        configuration option is set to False.
+        '''
+        if not asbool(config.get(DCAT_EXPOSE_SUBCATALOGS)):
+            return
+        cats = set(self.g.subjects(DCAT.dataset, dataset_ref))
```
amercader: Let's be explicit with variable names to avoid feline confusions and call this `catalogs` :)

ckanext/dcat/profiles.py
```diff
@@ -586,6 +619,25 @@ def parse_dataset(self, dataset_dict, dataset_ref):
         '''
         return dataset_dict
 
+    def _extract_catalog_dict(self, catalog_ref):
```
amercader: Considering that this data gets stored in every single dataset I think we should limit by default the fields that get stored. I understand that this data is used when serializing the catalog, but it seems really inefficient to store all these extras. I'd suggest storing only the remote catalog URI and maybe title to help with faceted search. Description, language and modified seem like secondary information, and the publisher JSON blob definitely seems like pushing it. Users can implement a custom profile if they want this extra fields or implement some lookup table.

ckanext/dcat/profiles.py
```diff
@@ -586,6 +619,25 @@ def parse_dataset(self, dataset_dict, dataset_ref):
...
+
+        out = []
+        sources = (('source_catalog_title', DCT.title,),
+                   ('source_catalog_description', DCT.description,),
+                   ('source_catalog_homepage', FOAF.homepage,),
+                   ('source_catalog_language', DCT.language,),
+                   ('source_catalog_modified', DCT.modified,),)
+
+        for key, predicate in sources:
+            val = self._object_value(catalog_ref, predicate)
```
amercader: if val:

ckanext/dcat/profiles.py
```diff
@@ -740,6 +793,12 @@ def parse_dataset(self, dataset_dict, dataset_ref):
         # License
         if 'license_id' not in dataset_dict:
             dataset_dict['license_id'] = self._license(dataset_ref)
+        
+        # Source Catalog
+        if catalog_src is not None:
```
amercader: if catalog_src is not None and asbool(config.get(DCAT_EXPOSE_SUBCATALOGS):

ckanext/dcat/processors.py
```diff
@@ -330,6 +333,70 @@ def serialize_catalog(self, catalog_dict=None, dataset_dicts=None,
...
+        source_uri = _get_from_extra('source_catalog_homepage')
+        if not source_uri:
+            return
+        
+        g = self.g
+        catalog_ref = URIRef(source_uri)
+
+        catalog_added = g.value(root_catalog_ref, DCT.hasPart) == catalog_ref
+
+        g.add((root_catalog_ref, DCT.hasPart, catalog_ref))
```
amercader: Shouldn't these two be inside `if not catalog_added`?

```
g.add((root_catalog_ref, DCT.hasPart, catalog_ref))
g.add((catalog_ref, RDF.type, DCAT.Catalog))
```

ckanext/dcat/profiles.py
etj: According to DCAT specifications, dct:description and dct;publisher are mandatory properties for the Catalog.

I agree for the customization of the other fields.

ckanext/dcat/processors.py
etj: Sure. Not spotted before because it was simply adding the same node over and over.

ckanext/dcat/profiles.py
etj: We are simply storing the source catalog info here. 
We may want to have them stored anyway, even if those info are not exposed during serialization.

amercader: I really want to make optional, defaulting to False, the storing of these fields. If you want you can add a second config option (`ckanext.dcat.store_subcatalogs` or something) that only triggers the storing. So the fields we'll be stored if `ckanext.dcat.expose_subcatalogs` or `ckanext.dcat.store_subcatalogs` are True

etj:
@amercader @cezio pls close this PR and refer to #101 instead: it's up to date with this PR, and includes some more fixes and improvements, and also follows the recommendations in review (such as  https://github.com/ckan/ckanext-dcat/pull/100#discussion_r149423850)

amercader:
Closed in favour of #101

Issue #75, opened 2016-12-01
Support for schema.org JSON-LD as a variation?


danbri:
There is a DCAT-based structure in schema.org for representing datasets, see http://schema.org/Dataset and nearby. We are using it at Google for dataset search, see https://developers.google.com/search/docs/data-types/datasets - would it be possible/appropriate to include support for this within your extension somehow, or is it better to have a separate addon for this? 

The syntax could be RDFa, Microdata or (probably best) JSON-LD, but should ideally be embedded in the main HTML page for each dataset. See http://developers.google.com/structured-data/testing-tool/ for Google's related Structured Data Testing Tool, which would show whether Google finds the markup.

amercader:
I think this is a brilliant idea, and certainly several people have expressed interest for schema.org integration. We definitely want to go for JSON-LD or whatever that can be embedded separately on the headers or wherever to not affect custom templates on other extensions.

A couple of questions (sorry is these sound silly):

* Is schema.org compatible with DCAT? I guess what I mean is whether the existing JSON-LD output (eg http://demo.ckan.org/dataset/newcastle-city-council-payments-over-500.jsonld) would be directly recognised by Google
* Just to confirm Google won't recognise linked markup (via `<link rel="alternate">`), it needs to be actually embedded on the page, right?

As for where the plugin should live, in this extension or elsewhere I don't really mind. Depending on the implementation is probably going to use some of the existing helpers to map standard CKAN fields to dcat/schema.org so having it here would be a good place to start.

danbri:
Thanks for the positive response! :) Schema.org's dataset structure is very similar to DCAT, but instead of using multiple Semantic Web vocabularies we re-used existing schema.org terms. If someone has a mainstream DCAT description of a dataset, it ought to be relatively straightforward to re-present it as a Schema.org Dataset. However there are various DCAT extensions / application profiles, DCAT-AP, -Geo, -Stat etc, which will need more discussion. I'm creating a W3C Community Group as a place for that interop work.

[...]

PR #108, opened 2018-01-16
[#75] Schema.org mapping

metaodi:
Add schema.org mapping (see #75 for details)

This PR provides:
* Mapping from DCAT to schema.org/Dataset
* New parameter `profiles` to the catalog and dataset endpoints to be able to change the profiles when requesting the metadata (this allows to server DCAT and schema.org on the same instance), example `/dataset/my-dataset.jsonld?profiles=schemaorg`

### DIFF
--- a/README.md
+++ b/README.md
@@ -804,7 +804,15 @@ To disable this behavior, you can set the following config value in your ini fil
 To add [structured data](https://developers.google.com/search/docs/guides/intro-structured-data) to dataset pages, activate the `structured_data` plugin in your ini file.
 
 By default this uses the `schemaorg` profile (see [profiles](#profiles)) to serialize the dataset to JSON-LD, which is then added to the dataset detail page.
-Example:
+To change the schema, you have to override the Jinja template block called `structured_data` in [`templates/package/read_base.html`](https://github.com/ckan/ckanext-dcat/blob/master/ckanext/dcat/templates/package/read_base.html) and call the template helper function with different parameters:
+
+    {% block structured_data %}
+      <script type="application/ld+json">
+      {{ h.structured_data(pkg.id, ['my_custom_schema'])|safe }}
+      </script>
+    {% endblock %}
+
+Example output of structured data in JSON-LD:
 
     < ... >
         <script type="application/ld+json">
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@ This extension provides plugins that allow CKAN to expose and consume metadata f
     - [Compatibility mode](#compatibility-mode)
 - [XML DCAT harvester (deprecated)](#xml-dcat-harvester-deprecated)
 - [Translation of fields](#translation-of-fields)
+- [Structured Data](#strucuted-data)
 - [Running the Tests](#running-the-tests)
 - [Acknowledgements](#acknowledgements)
 - [Copying and License](#copying-and-license)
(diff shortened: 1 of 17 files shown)


ckanext/dcat/controllers.py
```diff
@@ -32,10 +32,15 @@ def read_catalog(self, _format=None):
         if not _format:
             return HomeController().index()
 
+        _profiles = toolkit.request.params.get('profiles')
```
amercader: Does it really make sense to ask for more than one profile at the time? Can you give an example? IIRC profiles used inheritance to extend them.

amercader: Turns out I implemented it exactly like this at the lower level :upside_down_face: Ignore me!

metaodi:
I found a way to check if a helper function is defined.

In https://github.com/ckan/ckan/pull/3042 an excpetion was added to the helper dictionary, so it's no longer possible to simply use `{% if h.structured_data %}` in jinja, as this triggers an exception. But fortunately jinja allows us to convert `h` to a normal dict, which in turns allows us to check if a key exists:

```
{% set helper_dict = dict(h) %}
{% if helper.dict.structured_data %}
{{ h.structured_data() }}
{% endif %}
```

metaodi:
Unfortunately the approach I described above does not work in older versions of CKAN (<2.6), because earlier versions of jinja do not support the use of dict(). Therefore I added another template helper to check if a helper is available or not.

ckanext/dcat/utils.py
```diff
@@ -1,7 +1,16 @@
...
-from pylons import config
+from ckantoolkit import config, h
+
+try:
+    # CKAN >= 2.6
+    from ckan.exceptions import HelperError
+except ImportError:
+    # CKAN < 2.6
+    class HelperError(Exception):
+        pass
```
metaodi: I already added the `HelperError` to ckantoolkit to avoid this "hack" (https://github.com/ckan/ckantoolkit/pull/6)

metaodi:
I rebased this PR to master to fix the merge conflicts.

metaodi:
@amercader ping, can you review this PR?

amercader:
sorry @metaodi, I'm a bit swamped right now but it's definitely on my list. Will try to review it this week.

README.md
```diff
@@ -788,6 +799,88 @@ To disable this behavior, you can set the following config value in your ini fil
     ckanext.dcat.translate_keys = False
 
 
+## Structured data
+
+To add [structured data](https://developers.google.com/search/docs/guides/intro-structured-data) to dataset pages, activate the `structured_data` plugin in your ini file.
```
amercader: I took me quite a while to discover why I wasn't seeing the structured data json-ld after activating the `structured_data` plugin. Here we should mention that you need in `structured_data` in addition to the `dcat` plugin.

This basically means that if someone wants the schema.org support they also need to enable the links to RDF representation for datasets and catalog, as these are automatically added by the `dcat` plugin. This is probably fine, but we might want to add a config option to disable this for people that just want the schema.org support.

amercader:
This looks good @metaodi . Only one minor change in the README and we can merge.

metaodi:
@amercader I made the changes to the README and added the new config option for the RDF endpoints (+tests)

amercader:
Thanks @metaodi, fantastic work as usual

amercader:
This is worth a new release, I've released 0.0.7

PR #112, opened 2018-01-31
Remove dependency on ckanext-harvest

mattfullerton:
It doesn't seem right to assume that ckanext-harvest is available - sure, many use dcat for harvesting but you can also use it just for offering dcat. The only reason I can see for importing from ckanext-harvest is to support very, very old CKANs (https://github.com/ckan/ckanext-harvest/blob/master/ckanext/harvest/harvesters/base.py#L24).


TkTech:
👍

PR #113, opened 2018-02-20
If not format was given, use 'xml' as default

metaodi:


metaodi:
If you currently don't pass a format, `None` will be set as format, which causes a `PluginException` of the rdflib because None is not a recognized format.

E.g. https://demo.ckan.org/api/3/action/dcat_catalog_search?q=test causes this, but https://demo.ckan.org/api/3/action/dcat_catalog_search?q=test&format=json-ld works fine.

This PR sets 'xml' as the default, if no format was given.

PR #116, opened 2018-04-03
Json harvester reorg

davidread:
The gather/fetch/import_stage methods for the json harvester has been sitting in the base class DCATHarvester, also inherited by the DCATRDFHarvester, but that has it's own version of those methods, so it seems pretty safe to move them.

I added in the tests from #115 to check this works. This PR should probably be merged before that one and then I can sort out the conflicts.


amercader:
@davidread I haven't used the JSON harvester in a long time. Do you have recent experience with it, is it working as expected? When I wrote it ages ago I wasn't targeting data.json but if it works against it that would be awesome and we should document it.

davidread:
@amercader yes the json harvester seems to work ok. I'll just sort this other pull request #115 and then do a bit of docs.

PR #120, opened 2018-05-08
Use encodingFormat for schemaorg

metaodi:
The value for the field is either read from the 'format' field or the
mimetype field, whichever is not falsey.
Previously we set the fileType attribute as well, but this is a
non-existing schema.org attribute.

see #75


metaodi:
@danbri @amercader this is the PR to fix the encodingFormat

metaodi:
ping @amercader: any feedback to this PR?

amercader:
Sorry I missed this @metaodi

Issue #55, opened 2016-01-19
Support for multilingual RDF


amercader:
Right now, neither the parsers nor the serializers take multilingual metadata into account.

For instance given the following document, a random title among the three will be picked up during parsing time:

```
@prefix rdfs:    <http://www.w3.org/2000/01/rdf-schema#> .
@prefix dcat:    <http://www.w3.org/ns/dcat#> .
@prefix dct:     <http://purl.org/dc/terms/> .
@prefix xsd:     <http://www.w3.org/2001/XMLSchema#> .
@prefix rdf:     <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
@prefix skos:    <http://www.w3.org/2004/02/skos/core#> .

<http://data.london.gov.uk/dataset/Abandoned_Vehicles>
      a       dcat:Dataset ;
      dct:title "Abandoned Vehicles"@en ;
      dct:title "Vehículos Abandonados"@es ;
      adms:versionNotes "Some version notes"@en ;
      adms:versionNotes "Notas de la versión"@es ;

      ...
```

**Parsing**

The standard way of dealing with this seems to be to create metadata during the parsing that can be handled by [ckanext-fluent](https://github.com/ckan/ckanext-fluent) when creating or updating the datasets.
This essentially means storing a dict instead of a string, with the keys being the language codes:

```
{
    
    "version_notes": {
        "en": "Some version notes",
        "es": "Notas de la versión"
    }
    ...

}


``` 

For core fields like `title` or `notes`, we need to add an extra field suffixed with `_translated`:
 
```
    "title": "",
    "title_translated": {
        "en": "Abandoned Vehicles",
        "es": "Vehiculos Abandonados"
    }
    ...
```
 
TODO: what to put in `title`?

To support it we can proabably have a variant of [`_object_value`](https://github.com/ckan/ckanext-dcat/blob/72a4904c1c5f1f03055d76e2a4b9a695eb3fd453/ckanext/dcat/profiles.py#L101) that handles the lang tags and returns a dict accordingly (RDFLib will return a different triple for each language).

**Serializing**

Similarly, the serializing code could check the fields marked as multilingual to see if they are a string or a dict and create triples accordingly, proabably via a helper function.

--
Things to think about:

* Should this be the default or enabled via config option?
* This will probably require using ckanext-scheming as well, otherwise multilingual fields won't be properly stored.

amercader:
[...]

PR #124, opened 2018-07-17
add support for multilingual RDF

stefina:


seitenbau-govdata:
Hi @stefina,
we from the German open data portal are also interested in the multilingual support.

Is this the final implementation or is something missing? It seems that the actual change in the pull request changes the type of the expected default value if no value is in the graph to the given predicate. I think therefore some of the unit tests are failing, because the return value is different.

As @amercader mentioned in #55 there is a need to use ckanext-scheming for enabling proper storing of the multilingual fields in the database. A minimal example for the fields _title_ and _notes_ should be could be helpful to get started with it.

ckanext/dcat/profiles.py
```diff
@@ -104,17 +104,47 @@ def _object(self, subject, predicate):
...
+            elif multilang:
+                lang_dict[default_lang] = unicode(o)
+            else:
+                return unicode(o)
+        if multilang:
+            # when translation does not exist, create an empty one
+            for lang in get_langs():
+                if lang not in lang_dict:
+                    lang_dict[lang] = ''
+        return lang_dict
```
seitenbau-govdata: To keep the default behavior when nothing is found I suggest moving `return lang_dict` under `if multilang: ` and inserting `return ''` as default return value if `multilang=False` and nothing was found.

stefina: I agree, that makes sense. I updated my PR accordingly to that.

seitenbau-govdata: Thank's for your update!

stefina:
@seitenbau-govdata I added some documentation to make it clear how this is used. Would be good to hear your opinion on it.
Unfortunately the code-coverage-tests fail since I added documentation.
@amercader I think it makes sense to exclude the readme (or *.md-files altogether). What do you think? ;)

seitenbau-govdata:
@stefina It looks good to me. Maybe a hint would be helpful that the plugin ckanext-scheming is needed to store the values in several languages in the database properly. As I have understand the ckanext-scheming is required then, right?

ckanext/dcat/profiles.py
```diff
@@ -104,18 +104,49 @@ def _object(self, subject, predicate):
...
+                return unicode(o)
+        if multilang:
+            # when translation does not exist, create an empty one
+            for lang in get_langs():
+                if lang not in lang_dict:
+                    lang_dict[lang] = ''
+            return lang_dict
         return ''
 
+    def _add_multilang_value(self, subject, predicate, dataset_key, dataset_dict):  # noqa
```
amercader: It was not initially clear what this function did, would you mind adding a short docstring along the lines of "this function will add a new triple for each language defined in a multilingual field dict, eg ..."

Actually I'm wondering if the logic in this function could be incorporated to the existing helpers to add triples directly, eg [here](https://github.com/ckan/ckanext-dcat/blob/530c6d36c0889ea0301241939f82223c66c07405/ckanext/dcat/profiles.py#L530) so it is handled automatically on existing profiles without having to write your own and use `_add_multilang_value()`

ckanext/dcat/tests/test_base_profile.py
```diff
@@ -80,6 +80,33 @@ def test_object_value_not_found(self):
 
         eq_(value, '')
 
+    def test_object_value_multilang(self):
```
amercader: Can you add a small test for the other way around (ie from dict to graph)?

PR #129, opened 2018-08-08
Allow CI failures on master (unreleased CKAN version)

camfindlay:
This will allow a failure in the CI (so it appears as a pass for all other CKAN versions) on the master, given it's a ongonig working branch for CKAN core there will be times it breaks but this should not fail the build for known supported and released version of CKAN used with this module.

(Hopefully I've remembered my Travis config correctly! :D)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,3 +25,4 @@ before_install:
     - pip install codecov
 after_success:
     - codecov
+    
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,8 @@ env:
     - CKANVERSION=2.6
     - CKANVERSION=2.7
     - CKANVERSION=2.8
-
+allow_failures:
+    - env: CKANVERSION=master
 install:
     - bash bin/travis-build.bash
 services:


camfindlay:
Looks like 2.8 branch is also failing. Perhaps testing against specific stable version is a good idea?

amercader:
@camfindlay failures on 2.8 should be fixed in  986517c . Can you merge master and see if the build passes?

camfindlay:
@amercader - I'm not the maintainer 🤓 but I'll try triggering another CI run to at least check 2.8 passes. I still need a review to ensure I have my allow failures config set up correct in .travis.yml though. Figure master should be allowed to fail given it's where the coalface of major version development is going down. The minor branches however should pass otherwise it would show up a breaking change in core for a minor version (assuming SEMVER).

amercader:
@camfindlay sorry, I meant merge master on your PR.

I misunderstood your original intention, I though you only wanted to allow failures on master while https://github.com/ckan/ckan/pull/4385 was being fixed. Going forward I think I prefer to keep the config as is, ie failing against master, because that allows us to detect issues early and not when there's an actual release.

PR #130, opened 2018-08-13
Manage downloadURL or accessURL

### LINKED ISSUE #2: Manage downloadURL or accessURL
CKAN resources have just a single `url` property. We could rely on the resource type or format to decide if this is direct download.
For harvested datasets, we can probably store what kind of URL was defined and reuse it later on.

seitenbau-govdata:
Fixes #2.

Introduces a new property `access_url` for resources.

When parsing a dataset from RDF, the resource's `url` is set to `downloadURL` if present, with fallback to `accessURL`.

For serialization, `access_url` is used if available, and otherwise the old behavior is preserved (only add `accessURL` if it differs from `downloadURL`).


amercader:
Thanks @seitenbau-govdata! and sorry about taking a bit to merge

seitenbau-govdata:
Thank's for your work and time @amercader !

PR #132, opened 2018-08-29
Add RDF.type to resource checksum

seitenbau-govdata:
The RDF.type was missing for resource checksums.


metaodi:
Looks good to me.

PR #133, opened 2018-08-29
Make _object_value_int more robust by accepting decimals as well

seitenbau-govdata:
Parsing values containing dots led to value loss in `_object_value_int`.

This also parses floating point values to integers, which makes parsing more robust.

### DIFF
--- a/ckanext/dcat/profiles.py
+++ b/ckanext/dcat/profiles.py
@@ -128,7 +128,7 @@ class RDFProfile(object):
         object_value = self._object_value(subject, predicate)
         if object_value:
             try:
-                return int(object_value)
+                return int(float(object_value))
             except ValueError:
                 pass
         return None
--- a/ckanext/dcat/tests/test_base_profile.py
+++ b/ckanext/dcat/tests/test_base_profile.py
@@ -94,6 +94,20 @@ class TestBaseRDFProfile(object):
         assert isinstance(value, int)
         eq_(value, 23)
 
+    def test_object_int_decimal(self):
+
+        p = RDFProfile(_default_graph())
+
+        p.g.add((URIRef('http://example.org/datasets/1'),
+                 TEST.some_number,
+                 Literal('23.0')))
+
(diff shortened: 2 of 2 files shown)


metaodi:
What is the use case of this? This might lead to unexpected behavior for users, since values are not rounded, i.e. the literal "3.8" will be saved as the integer 3. As far as I can see this method is currently used for the `byteSize`. I'm not entirely sure if it's better to silently lose the value (since it's not a valid integer) or to just "cut" it, to make it valid.

seitenbau-govdata:
Thank's for your feedback, @metaodi .

Some of our harvested datasets contain values like 16.0 in `byteSize`.
As looking for a `.0` is more complex than simply calling float, we decided to implement it like that.

You are right, results could be unexpected for users. But silently loosing the value is worse in our opinion.

So we suggest one of those options:
* round the value
* cut the value (as implemented right now)
* when the value is a float, only keep it if it ends with `.0`

seitenbau-govdata:
@metaodi Please let me know what your preferred choice for the implementation is. If you think it's completely not wanted we can also close the pull request.

metaodi:
@seitenbau-govdata I think it's fine like it is, cutting is better than losing. And rounding the value might not be expected as well.

seitenbau-govdata:
@metaodi Ok, great. Who is responsible merging the PR? You or @amercader ? Of course, I'm always interested in @amercader opinion. :-)

metaodi:
@seitenbau-govdata I'm not (yet 😉 ) a maintainer of this repo. So @amercader is responsible 😊

seitenbau-govdata:
Ah, ok. Then I confused it with the repo https://github.com/ckan/ckanext-harvest. 😊 But there your'e a maintainer.

metaodi:
@amercader sure, I'd like to help and I'm anyway reviewing PRs

amercader:
@metaodi fantastic thanks! You should now have write permissions :sunglasses:

PR #134, opened 2018-08-29
Improve mailto handling and add generic value modifier to _add_triple_from_dict

seitenbau-govdata:
The generic modifier allows to provide a function which updates the value before it is added to the graph from the dict.

Also, this improves handling of mail addresses:

* Use URIRef instead of Literals for email addresses
* ensure that `mailto:` is prepended to the address

### DIFF
--- a/ckanext/dcat/profiles.py
+++ b/ckanext/dcat/profiles.py
@@ -475,8 +475,8 @@ class RDFProfile(object):
                               fallbacks=None,
                               list_value=False,
                               date_value=False,
-                              value_modifier=None,
-                              _type=Literal):
+                              _type=Literal,
+                              value_modifier=None):
         '''
         Adds a new triple to the graph with the provided parameters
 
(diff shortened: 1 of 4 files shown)


ckanext/dcat/profiles.py
```diff
@@ -471,15 +475,19 @@ def _add_triple_from_dict(self, _dict, subject, predicate, key,
                               fallbacks=None,
                               list_value=False,
                               date_value=False,
+                              value_modifier=None,
```
metaodi: I would add the new parameter at the end, maybe someone is already using `_add_triple_from_dict` in their code, so this would break it.

amercader:
This looks good @seitenbau-govdata. @metaodi raises a good point, so I think it's safer to add the new keyword param at the end.

seitenbau-govdata:
@amercader and @metaodi Good point! I agree. I have pushed a new commit with the change.

amercader:
Thanks @seitenbau-govdata

PR #138, opened 2018-09-06
Encode URIRef and remove white-spaces

seitenbau-govdata:
The pull request adds an careful URIRef quoting for URIs with http(s) and white-space removal for URIs without http(s), e.g. mailto.

It mainly solves an internal server error when serializing the graph. If there are some datasets in the database with URIs containing white-spaces, it is not possible to serialize the graph which returns into an internal server error.
Furthermore it solves the problem with incorrect encoded URIs which can result in a problem in a remote system when it is importing the graph.

### DIFF
--- a/ckanext/dcat/profiles.py
+++ b/ckanext/dcat/profiles.py
@@ -1,6 +1,8 @@
 import datetime
 import json
 
+from urllib import quote
+
 from dateutil.parser import parse as parse_date
 
 from ckantoolkit import config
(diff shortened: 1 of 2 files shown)


ckanext/dcat/profiles.py
```diff
@@ -45,6 +47,34 @@
...
+        for c in quotechars:
+            value = value.replace(c, quote(c))
+        return value
+
+    def __new__(cls, value):
+        if isinstance(value, basestring):
+            if value.startswith('http'):
+                value = CleanedURIRef._careful_quote(value)
+            else:
+                value = value.replace(' ', '')
```
metaodi: Why do you remove spaces? In the unit test you have the example "mailto:with space@example.com " is converted to "mailto:withspace@example.com". IMHO this is wrong and just buries the error. To remove the spaces from either end of the string (using e.g. `strip()`) could be a solution. But we shouldn't change the data, this should lead to an error and the data should be corrected. Otherwise withspace@example.com might receive emails that were actually meant to with_space@example.com.

amercader: That sounds sensible to me

PR #139, opened 2018-09-12
Add missing schema.org fields

metaodi:
This PR adds the following fields:

- url of the dataset (SCHEMA.url)
- group URLs to SCHEMA.genre
- license to SCHEMA.license
- information about the catalog as a SCHEMA.DataCatalog

Additionally a new method `additional_fields` is added, which makes it easy in a subclass to add their own fields without having to override one of the "internal methods" of the provided SchemaOrgProfile class

### DIFF
--- a/ckanext/dcat/profiles.py
+++ b/ckanext/dcat/profiles.py
@@ -1251,6 +1251,7 @@ class SchemaOrgProfile(RDFProfile):
             ('version', SCHEMA.version, ['dcat_version'], Literal),
             ('issued', SCHEMA.datePublished, ['metadata_created'], Literal),
             ('modified', SCHEMA.dateModified, ['metadata_modified'], Literal),
+            ('license', SCHEMA.license, ['license_url', 'license_title'], Literal),
         ]
         self._add_triples_from_dict(dataset_dict, dataset_ref, items)
 
--- a/ckanext/dcat/profiles.py
+++ b/ckanext/dcat/profiles.py
@@ -1258,6 +1258,13 @@ class SchemaOrgProfile(RDFProfile):
 
         self._add_date_triples_from_dict(dataset_dict, dataset_ref, items)
 
+        # Dataset URL
+        dataset_url = url_for(controller='package',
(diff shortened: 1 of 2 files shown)


metaodi:
@amercader maybe you can have a look at this. When I was doing an integration lately, I noticed these fields were missing and they IMHO make sense to provide in the basic implementation. And by adding `additional_fields` we can provide an extension point for everybody, who creates a subclass of `SchemaOrgProfile` to provide a custom mapping for their schema.

metaodi:
@amercader do you have feedback for me? Or can you merge this?

amercader:
This looks good overall @metaodi. The only thing I'm not 100% sold on is using groups as schema.genre. Is there an existing recommendation for using this property for topics, themes, etc

@danbri can you think of a better property?

@dan

metaodi:
@amercader I'm not 100% sure it's correct. But by looking at the example on https://schema.org/genre, where it is used to describe "Late Renaissance" (using the URL http://vocab.getty.edu/aat/300021143), I think it fits pretty good with the CKAN groups.
The description of the field on schema.org is "Genre of the creative work, broadcast channel or group", which imho indicates that CKAN groups would be a good match.

amercader:
@metaodi just revisiting this. I don't see why we can't use multiple `SCHEMA.about` properties for groups, instead of `SCHEMA.genre`. The [docs](https://schema.org/about) say "The subject matter of the content", which I think fits CKAN groups quite well.

If you are happy with this we can merge.

metaodi:
@amercader `SCHEMA.about` seems like a good fit, since it's a `Thing` we can provide a URL and a name/description of the group, thus represent the group even better. I'm currently on vacation, but happy to change that when get to it.

amercader:
@metaodi I wanted to release a new version while I had some time available so I went ahead and did the change myself (https://github.com/ckan/ckanext-dcat/commit/67c2a5e5b99ac988b36eb466b190b43d4e1b5ab2). Thanks for the PR

PR #140, opened 2018-10-12
Assign URIRef or Literal types based on content

seitenbau-govdata:
This introduces a factory class which uses the node type URIRef if the value seems to be an HTTP(S) URL, which is more accurate in that case. Otherwise, Literal is used.

### DIFF
--- a/ckanext/dcat/profiles.py
+++ b/ckanext/dcat/profiles.py
@@ -51,8 +51,23 @@ namespaces = {
 
 PREFIX_MAILTO = u'mailto:'
 
+class URIRefOrLiteral(object):
+    '''Helper which creates an URIRef if the value appears to be an http URL,
+    or a Literal otherwise. URIRefs are also cleaned using CleanedURIRef.
+
+    Like CleanedURIRef, this is a factory class.
+    '''
+    def __new__(cls, value):
+        stripped_value = value.strip()
+        if (isinstance(value, basestring) and (stripped_value.startswith("http://")
+                                               or stripped_value.startswith("https://"))):
+            return CleanedURIRef(value)
+        else:
+            return Literal(value)
+
+
 class CleanedURIRef(object):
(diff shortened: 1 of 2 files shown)


seitenbau-govdata:
@amercader and @metaodi Please let me know your thoughts about the pull request. Do you have some feedback or merge this?

ckanext/dcat/profiles.py
```diff
@@ -51,8 +51,23 @@
...
+class URIRefOrLiteral(object):
+    '''Helper which creates an URIRef if the value appears to be an http URL,
+    or a Literal otherwise. URIRefs are also cleaned using CleanedURIRef.
+
+    Like CleanedURIRef, this is a factory class.
+    '''
+    def __new__(cls, value):
+        stripped_value = value.strip()
+        if (isinstance(value, basestring) and (stripped_value.startswith("http://")
+                                               or stripped_value.startswith("https://"))):
```
metaodi: I think this check needs to verify if the result is a valid URL. According to the rdflib docs, it will happily create a URIRef of a URI containing invalid characters, but it might fail when serializing it. I'd say this factory should create a Literal for invalid URIs.

seitenbau-govdata: Good point @metaodi, thank you! In rdflib, only [some characters are checked](https://github.com/RDFLib/rdflib/blob/f2d893145f6fba3513d5e169b964554451c5eee9/rdflib/term.py#L253) at the moment, which are now all encoded in CleanedURIRef.

However, to handle possible changes in rdflib, serialization is tried anyway and Literal gets now returned in case something fails.

seitenbau-govdata: @metaodi Do you think the PR is ready to merge now?

amercader:
BTW 2.7 failures should now be fixed on master

PR #144, opened 2018-12-18
Improved dct:format and dcat:mediaType handling

seitenbau-govdata:
This correctly parses URI values for dct:format (which were silently lost before) and improves serializing of dcat:mediaType and dct:format.

If a value in `resource['format']` can be interpreted as IANA media type, it is output as dcat:mediaType, following the standard where dcat:mediaType is the more specialized field and subset of dct:format.


amercader:
This looks good, thanks @seitenbau-govdata

seitenbau-govdata:
@amercader You're welcome, and thanks for merging so fast!

PR #145, opened 2019-01-16
Updating the URLs to dataportals.org

LeoFrachet:
AFAICS `spec.datacatalogs.org` is now `spec.dataportals.org`.

This proposal aims to update the URLs to the new endpoints.

Thanks


amercader:
Thanks @LeoFrachet!

PR #149, opened 2019-03-14
Handle import stage errors

kentsanggds:
## What

Our fetch process had stopped because of Validation Errors which were not properly handled in the `import_stage` this would eventually lead to the process dying after a number of retries. The end user would be unaware of any problems with the data as the status of the job is not updated.

## How to review

2 tests have been added to demonstrate the new behaviour which fixes the issue: 

`test_harvest_does_not_create_with_invalid_tags` -  demonstrates that it will not create the dataset with invalid data.

`test_import_invalid_tags` - checks that a Validation Error is handled and would be stored in an error log which the end user will have visibility of.

### DIFF
--- a/ckanext/dcat/harvesters/_json.py
+++ b/ckanext/dcat/harvesters/_json.py
@@ -272,15 +272,19 @@ class DCATJSONHarvester(DCATHarvester):
                 'SET CONSTRAINTS harvest_object_package_id_fkey DEFERRED')
             model.Session.flush()
 
-            package_id = \
-                p.toolkit.get_action('package_create')(context, package_dict)
-            log.info('Created dataset with id %s', package_id)
-
         elif status == 'change':
             package_dict['id'] = harvest_object.package_id
-            package_id = \
-                p.toolkit.get_action('package_update')(context, package_dict)
-            log.info('Updated dataset with id %s', package_id)
+
+        if status in ['new', 'change']:
+            try:
+                action = 'package_create' if status == 'new' else 'package_update'
+                message_status = 'Created' if status == 'new' else 'Updated'
+
+                package_id = p.toolkit.get_action(action)(context, package_dict)
+                log.info('%s dataset with id %s', message_status, package_id)
+            except p.toolkit.ValidationError, e:
+                self._save_object_error('Validation Error: %s' % str(e), harvest_object, 'Import')
+                return False
 
         model.Session.commit()
 
--- a/ckanext/dcat/harvesters/_json.py
+++ b/ckanext/dcat/harvesters/_json.py
@@ -1,6 +1,7 @@
 import json
 import logging
 from hashlib import sha1
+import traceback
 import uuid
 
 from ckan import model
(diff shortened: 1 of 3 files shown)


kentsanggds:
@amercader any thoughts on this PR?

kentsanggds:
@metaodi any chance that you can have a look at this PR?

ckanext/dcat/tests/test_json_harvester.py
```diff
@@ -238,3 +270,55 @@ def test_not_copied_because_completely_different(self):
...
+
+class TestImportStage:
+
+    @classmethod
+    def setup_class(cls):
+        h.reset_db()
+
+    class MockHarvestObject:
+        guid = 'test_guid'
+        content = TestDCATJSONHarvestFunctional.json_content_invalid_tags
```
metaodi: Maybe create a constructor and pass the content when creating the object. For now we only have this one test that makes use of this, but it might be handy for other tests as well. WDYT?

ckanext/dcat/harvesters/_json.py
```diff
@@ -272,15 +272,19 @@ def import_stage(self, harvest_object):
...
+        if status in ['new', 'change']:
+            try:
+                action = 'package_create' if status == 'new' else 'package_update'
+                message_status = 'Created' if status == 'new' else 'Updated'
+
+                package_id = p.toolkit.get_action(action)(context, package_dict)
+                log.info('%s dataset with id %s', message_status, package_id)
+            except p.toolkit.ValidationError, e:
+                self._save_object_error('Validation Error: %s' % str(e), harvest_object, 'Import')
+                return False
```
metaodi: I think checking for a ValidationError is super and helps to get a proper error message. May a [general try/except just like we have in the RDF harvester](https://github.com/ckan/ckanext-dcat/blob/f15cdfa6a8a9df627107e86e7ee2d98928c4a7b6/ckanext/dcat/harvesters/rdf.py#L385-L390) could be added to make sure we always get a proper error message.

ckanext/dcat/tests/test_json_harvester.py
kentsanggds: It's possible but I also think that it's not worth over engineering things until you actually need them, so if another test comes along that also wants to use this MockHarvestObject and to pass in different content then its worth making it more generic.

ckanext/dcat/harvesters/_json.py
kentsanggds: Generally it's probably better to be more narrow when catching exceptions so that errors are not hidden but seeing that there's existing code in `dcat` that handles it this way I can update the code to make it consistent.

metaodi: Yes to be narrow and specific is the pythonic way. But the problem with harvesters is, that the error is actually hidden if you don't catch it. The daemon might fail, but usually restarted. So to get an error on the UI, an object_error should be saved.

ckanext/dcat/tests/test_json_harvester.py
metaodi: I fully agree

ckanext/dcat/harvesters/_json.py
kentsanggds: Another approach would be to have a catch all in `ckanext-harvest` instead in case the plugins don't handle the more narrow errors. This would put the onus on the plugins to handle errors they are aware of but if all else fails it won't kill the fetch process as it would be caught by the harvest plugin. 

For now I'll update the code for consistency.

kentsanggds: I've updated the code to include a catch all exception handling but also have left the existing one in place as I think the existing error message is much clearer for the end user to act on than the catch all message which includes a traceback.

kentsanggds: @metaodi I've updated the code, let me know what you think

metaodi:
Looks good to me.

As for your suggestion to catch these errors already in ckanext-harvest: I agree, this would be a good idea. I'll created an issue to track this: https://github.com/ckan/ckanext-harvest/issues/362

PR #152, opened 2019-04-16
Pass `q` and `fq` parameters in catalog endpoint

### LINKED ISSUE #151: Add `q` and `fq` parameter to catalog endpoint
The action `dcat_catalog_show` currently allows to pass the `q` and `fq` parameters to limit the catalog output with a search query (i.e. only certain tags or organizations). Unfortunately the catalog endpoint (`/catalog.(xml|rdf|ttl)`) does not expose these parameters (i.e. they are not passed to the `dcat_catalog_show` action call).

I propose to change that, so that when harvesting an instance using it's catalog endpoint, it's possible to limit the returned datasets based on a search query/filter query.

metaodi:
Fixes #151

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,7 +187,7 @@ It's possible to specify the profile(s) to use for the serialization using the `
 
 http://demo.ckan.org/catalog.xml?profiles=euro_dcat_ap,sweden_dcat_ap
 
-To filter the output, the catalog endpoint supports the `q` and `fq` parameters to specify a search query or filter query:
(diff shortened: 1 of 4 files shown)


metaodi:
@amercader ping :smiley:

metaodi:
@amercader do you have time to review this? I already tested this on our integration environment. It works for me :wink:

amercader:
Sorry @metaodi, too many open fronts. Thanks for this!

PR #154, opened 2019-06-14
Include templates in package

seitenbau-govdata:
Includes templates in the package.


seitenbau-govdata:
@amercader Thanks for the quick merge!

PR #156, opened 2019-06-25
[#155] Ignore auth in internal search call

### LINKED ISSUE #155: CKAN 2.7.2 Error - <class 'ckan.logic.NotAuthorized'>:
When i open the Catalog.rdf i get:
Error - <class 'ckan.logic.NotAuthorized'>:
what could be the issue/which infos do u need?

dataset/nameofdataset.rdf works tho

amercader:
Fixes #155

### DIFF
--- a/ckanext/dcat/profiles.py
+++ b/ckanext/dcat/profiles.py
@@ -647,8 +647,7 @@ class RDFProfile(object):
         found.
         '''
         context = {
-            'user': toolkit.get_action('get_site_user')(
-                {'ignore_auth': True})['name']
+            'ignore_auth': True
         }
         result = toolkit.get_action('package_search')(context, {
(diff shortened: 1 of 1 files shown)


metaodi:
The build error on master seems to be related to https://github.com/ckan/ckan/pull/4853 and are caused by [code in ckanext-harvest](https://github.com/ckan/ckanext-harvest/blob/3a72337f1e619bf9ea3221037ca86615ec22ae2f/ckanext/harvest/logic/dictization.py#L63-L79).

Note sure if simply adding `text('error_count desc')` instead of `error_count desc` already solves the problem, but it's anyway not related to this PR or repo.

amercader:
Thanks for having a look @metaodi !

PR #158, opened 2019-07-09
Also support URIRef for dct:language

seitenbau-govdata:
If the dct:language is an URI (e.g. if it refers to a code list), output it as URIRef in the EuropeanDCATAPProfile.

### DIFF
--- a/ckanext/dcat/tests/test_euro_dcatap_profile_serialize.py
+++ b/ckanext/dcat/tests/test_euro_dcatap_profile_serialize.py
@@ -108,7 +108,7 @@ class TestEuroDCATAPProfileSerializeDataset(BaseSerializeTest):
                 {'key': 'alternate_identifier', 'value': '[\"xyz\", \"abc\"]'},
                 {'key': 'version_notes', 'value': 'This is a beta version'},
                 {'key': 'frequency', 'value': 'monthly'},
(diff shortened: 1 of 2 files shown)


ckanext/dcat/tests/test_euro_dcatap_profile_serialize.py
```diff
@@ -606,13 +610,17 @@ def test_distribution_fields(self):
         # List
         for item in [
             ('documentation', FOAF.page, URIRef),
-            ('language', DCT.language, Literal),
+            ('language', DCT.language, [Literal, Literal, URIRef]),
```
metaodi: This should probably be `[Literal, URIRef]`

seitenbau-govdata: `Literal` is contained twice because a type is assigned to each item in the test value list. As the list (line 575, the one above in the diff) contains three items, the data type list contains three items as well.

metaodi: Okay got it. Thanks for explaining.

seitenbau-govdata: You're welcome!

PR #159, opened 2019-07-09
Support JSON-LD catalogs with @graph

seitenbau-govdata:
Adds support for parsing JSON-LD catalogs containing `@graph` elements.

Links to open issues in rdflib-jsonld relating this problem:
- https://github.com/RDFLib/rdflib-jsonld/issues/40
- https://github.com/RDFLib/rdflib-jsonld/issues/53

The suggested "solution" is to use `rdflib.ConjunctiveGraph()` instead of `rdflib.Graph()`.

### DIFF
--- a/ckanext/dcat/processors.py
+++ b/ckanext/dcat/processors.py
@@ -67,7 +67,7 @@ class RDFProcessor(object):
                 config.get(COMPAT_MODE_CONFIG_OPTION, False))
         self.compatibility_mode = compatibility_mode
 
-        self.g = rdflib.Graph()
+        self.g = rdflib.ConjunctiveGraph()
 
     def _load_profiles(self, profile_names):
         '''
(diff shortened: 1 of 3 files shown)


metaodi:
@seitenbau-govdata is there any downside in using `ConjunctiveGraph`? So this should work for all our cases? I see that all tests pass, so I guess everything is fine.

seitenbau-govdata:
@metaodi We have had the same thoughts. Ultimately we don't know it if something special is not working withe the `ConjunctiveGraph`. So, we have to trust the implemented tests that all relevant use cases are covered.

seitenbau-govdata:
@metaodi What do you think about the pull request? Maybe it's helpful to have another opinion.
@amercader Please let us know your view about the pull request.

metaodi:
I like it and think it's a good solution. I'd say let's risk it and test in all our installations, if something is broken, we can still revert if needed.

amercader:
Tests with the alternative JSON-LD syntax seem to pass, so let's merge it

seitenbau-govdata:
Thanks @metaodi  and @amercader !

PR #161, opened 2019-07-29
Extract read datasets from db to make it re-usable

seitenbau-govdata:
Moves the code for reading datasets by the guid into a separate method to make it possible to re-use the method instead of using the method `_get_existing_datasets` (calls package_show additionally which is unnecessary in our case) or copy paste the code in our harvester implementation.


metaodi:
LGTM

seitenbau-govdata:
Thanks @metaodi !

PR #167, opened 2019-11-22
Fix for Travis build, now requirements.txt on master is for python3

davidread:


davidread:
Merged with consent from @smotornyuk and @tino097

PR #174, opened 2020-02-04
Python 3 support

amercader:
Includes the great work from @smotornyuk on #171, plus

* fixes
* New pytest based test suite
* Updated Travis setup

Tests on CKAN 2.9 / Python 3 fail because ckanext-harvest has not been ported yet.


amercader:
@metaodi @seitenbau-govdata I'm going to merge this as I need to use it as a reference for other extension py3 migrations but please flag any comments you find in the future.

You will notice that there are two test suites, the one in the `tests/nose` folder is the one you know, and it targets CKAN <= 2.8. The new default one is the one in the `tests` folder, based on pytest. From now on we'll need to add tests on both suites (we might work on a unified backwards compatible suite in the future)

seitenbau-govdata:
@amercader I know I'm a little bit late...😃 Thanks a lot. Great work!

amercader:
Thanks @seitenbau-govdata !

PR #175, opened 2020-02-07
Fix format normalization configuration

seitenbau-govdata:
We wanted to deactivate the ckan format normalization by config. But we are surprised that the config parameter `ckanext.dcat.normalize_ckan_format` seems to be ignored. We invested some time and found out the reason.
The reason is that the config parameter was read from the config file but was interpreted as string what is still the default behavior for an ini file. So, the value `False` in the config file was interpreted as string `'False'` and not as boolean what will lead to the situation that the expression `and normalize_ckan_format` was always True, because a not empty string is interpreted as True.

The pull request adds the conversion of the param to type boolean.

### DIFF
--- a/ckanext/dcat/profiles.py
+++ b/ckanext/dcat/profiles.py
@@ -947,8 +947,8 @@ class EuropeanDCATAPProfile(RDFProfile):
                     resource_dict[key] = json.dumps(values)
 
             # Format and media type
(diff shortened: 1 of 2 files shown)


seitenbau-govdata:
@amercader or @metaodi Maybe one of you will find time to have a look at this PR? Thanks in advance!

PR #176, opened 2020-02-07
Introduce the possibility to modify package update/create schema

seitenbau-govdata:
We want to deactivate the email validation for the fields `author_email` and `maintainer_email`, because we would not like to reject the import of a dataset if the email address is invalid.

The pull request introduces the possibility to modify/overwrite the update/create package schema used in the harvest process updating/creating datasets. This can be done by implementing the new interface methods `update_package_schema_for_create` and `update_package_schema_for_update`.

### DIFF
--- a/ckanext/dcat/harvesters/rdf.py
+++ b/ckanext/dcat/harvesters/rdf.py
@@ -308,7 +308,13 @@ class DCATRDFHarvester(DCATHarvester):
         existing_dataset = self._get_existing_dataset(harvest_object.guid)
 
         try:
+            package_plugin = lib_plugins.lookup_package_plugin(dataset.get('type', None))
             if existing_dataset:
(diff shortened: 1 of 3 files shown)


amercader:
This sounds useful, I might steal it for other harvesters! My only concern is that we will now have two ways in which harvesters can be extended, inheriting from this base class and overriding the `_get_create/update_package_schema` methods or implementing the `IDCATRDFHarvester` interface (which implements `after/before_update/create`).

Perhaps for consistency we can add new methods to that interface?

seitenbau-govdata:
Good point, @amercader ! We will try to change it and make use of the `IDCATRDFHarvester` interface.

seitenbau-govdata:
@amercader I have changed the pull request to using interface methods. Please review again. :-)

Issue #179, opened 2020-03-10
Unable to login after installing


segalindoa:
Installation goes well, but start getting an 500 internal server error when I try to login.

If i remove the related dcat values from ckan.plugins parameters in .ini file, login starts working fine again.

This is what i fnd in the log:

mod_wsgi (pid=11135): Exception occurred processing WSGI script '/etc/ckan/default/apache.wsgi'.
IOError: failed to write data
mod_wsgi (pid=11135): Exception occurred processing WSGI script '/etc/ckan/default/apache.wsgi'.
...
TypeError: expected byte string object, value of type unicode found

Thanks in advance for your help

segalindoa:
thanks a lot!
Up an running.

PR #181, opened 2020-03-25
Fixed login Internal server error

styhar:
This change fixes login Internal server error mentioned here:
https://github.com/ckan/ckanext-dcat/issues/179


smotornyuk:
Tested and verified. I think, it would be nice to remove this `from past.builtins import basestr` and use six.str_type instead

Issue #188, opened 2021-01-12
Harvesting a dcat catalog raises error


quaxsze:
Harvesting a DCAT catalog with python 3 version raises an `UnicodeDecodeError`

`'utf-8' codec can't decode byte 0xc3 in position 1023: unexpected end of data`

```
for chunk in r.iter_content(chunk_size=self.CHUNK_SIZE):
    if six.PY2:
        content = content + chunk
    else:
        content = content + chunk.decode('utf8')
```
This happens because trying to decode each chunk leads to decoding truncated characters.

Content should be decoded once fully downloaded.

PR #189, opened 2021-01-12
Fix python 3 harvest's encoding error

quaxsze:
fix https://github.com/ckan/ckanext-dcat/issues/188


quaxsze:
CI fails because of internal permissions

amercader:
Thanks @quaxsze, makes sense

PR #193, opened 2021-03-02
Add a response passthru for solr_url config

kentsanggds:
## What

Without the response passthru the harvest tests will fail with an external redirect detected.

This will allow tests to pass on a docker dev stack.


amercader:
Thanks @kentsanggds , done in https://github.com/ckan/ckanext-dcat/commit/fc714f1084e61c70b2b475c0de518267597b56dd

Issue #192, opened 2021-02-16
beautifulsoup4 v4.3.2 not supported by Python v3.5+


alycejenni:
Is there a specific reason that beautifulsoup4 is set to v4.3.2 in dev_requirements.txt, or has there just not been a reason to update it?
https://github.com/ckan/ckanext-dcat/blob/6b7ec505f303fb18e0eebcebf67130d36b3dca82/dev-requirements.txt#L1

Using the default or `html5lib` parser with bs4 in Python 3.5+ results in import errors for `HtmlParseError`, since it was [removed in 3.5](https://bugs.python.org/issue15114). Updating to bs4 v4.4.0+ fixes it.

(I'm not actually getting errors from ckanext-dcat by itself, just when it's used with other plugins that also use bs4!)

amercader:
@alycejenni there's no reason to not update it. TBH I think we should be fine leaving the dev requirements unpinned. 
And actually I don't think ckanext-dcat actually uses BS for anything does it? If so it can just be removed

PR #195, opened 2021-03-17
Remove beautifulsoup4 from dev_requirements.txt

alycejenni:
It doesn't appear to be used anywhere and causes errors when used with Python 3.5+ (see ckan/ckanext-dcat#192).

(duplicate of #194 because I renamed the branch and accidentally closed the original PR, sorry :upside_down_face: )


amercader:
Thanks @alycejenni

PR #196, opened 2021-04-14
Introduce interface method `after_parsing`

seitenbau-govdata:
Adds a new harvester interface method which is called just after the content from the remote RDF file has been parsed.

It allows to interact with the RDF parser instance, which can be useful to work with the graph and put it to other stores, e.g. a triple store.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased](https://github.com/ckan/ckanext-dcat/compare/v1.1.0...HEAD)
 
 * TBD
+* Introduce new interface method `after_parsing`
 
 ## [v1.1.1](https://github.com/ckan/ckanext-dcat/compare/v1.1.0...v1.1.1) - 2021-03-17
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## [Unreleased](https://github.com/ckan/ckanext-dcat/compare/v1.1.0...HEAD)
 
 * TBD
+* Introduce new interface method `after_parsing`
```
amercader, line 6:
```suggestion
* Introduce new interface method `after_parsing` (#196)
```

amercader (approved):
Looks good @seitenbau-govdata! One tiny comment and you can merge this yourself

CHANGELOG.md
seitenbau-govdata: @amercader Thanks for the fast review and the suggested change. I will push a new commit updating the changelog with some additional minor changes, e.g. changing the comparison from the latest 1.1.1 to the current master.

PR #203, opened 2021-11-05
Update changelog and setup.py for version 1.1.3

seitenbau-govdata:
As there have been some changes since the last release, we would like to release a new version 1.1.3. What do you think, @amercader @metaodi?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,13 @@
 # Changelog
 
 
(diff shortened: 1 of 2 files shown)


amercader:
If all tests are green, let's gooo :rocket: 
@seitenbau-govdata you should have access to merge and push new tags

seitenbau-govdata:
Thank you for the quick response. A new tag 1.1.3 has been created.

PR #208, opened 2022-04-13
Support CKAN 2.10

avdata99:
Adding support for CKAN 2.10.

Related to [CKAN#6787](https://github.com/ckan/ckan/issues/6787)

CKAN master + harvest + DCAT looks good

image

This extensions requires `ckanext-harvest` and [Harvest#496](https://github.com/ckan/ckanext-harvest/pull/496) to be merged
(I sent a PR to that PR to update it [here](https://github.com/salsadigitalauorg/ckanext-harvest/pull/1/files))

With those PR merged, locally, I get all tests ok

image

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,7 +4,7 @@ jobs:
   lint:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - uses: actions/setup-python@v2
         with:
           python-version: '3.6'
@@ -44,9 +44,7 @@ jobs:
       CKAN_REDIS_URL: redis://redis:6379/1
 
     steps:
-    - name: adding github workspace as safe directory
-      run: git config --global --add safe.directory $GITHUB_WORKSPACE
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
     - name: Install requirements
       run: |
         pip install -r requirements.txt
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,7 +17,7 @@ jobs:
     needs: lint
     strategy:
       matrix:
-        ckan-version: [2.9, 2.9-py2, 2.8, 2.7]
+        ckan-version: ["2.10", 2.9, 2.9-py2, 2.8, 2.7]
       fail-fast: false
 
     name: CKAN ${{ matrix.ckan-version }}
@@ -26,7 +26,7 @@ jobs:
       image: openknowledge/ckan-dev:${{ matrix.ckan-version }}
     services:
       solr:
-        image: ckan/ckan-solr-dev:${{ matrix.ckan-version }}
+        image: ckan/ckan-solr:${{ matrix.ckan-version }}
       postgres:
         image: ckan/ckan-postgres-dev:${{ matrix.ckan-version }}
         env:
@@ -44,6 +44,8 @@ jobs:
       CKAN_REDIS_URL: redis://redis:6379/1
 
     steps:
+    - name: adding github workspace as safe directory
+      run: git config --global --add safe.directory $GITHUB_WORKSPACE
     - uses: actions/checkout@v2
     - name: Install requirements
       run: |
--- a/ckanext/dcat/profiles.py
+++ b/ckanext/dcat/profiles.py
@@ -20,8 +20,7 @@ from geomet import wkt, InvalidGeoJSONException
 from ckan.model.license import LicenseRegister
 from ckan.plugins import toolkit
 from ckan.lib.munge import munge_tag
-from ckan.lib.helpers import url_for
-
+from ckanext.dcat.urls import url_for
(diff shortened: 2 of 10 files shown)


ckanext/dcat/profiles.py
```diff
@@ -459,7 +457,8 @@ def _access_rights(self, subject, predicate):
             if isinstance(obj, BNode) and self._object(obj, RDF.type) == DCT.RightsStatement:
                 result = self._object_value(obj, RDFS.label)
             elif isinstance(obj, Literal) or isinstance(obj, URIRef):
-                result = six.text_type(obj)
+                # unicode_safe not include Literal or URIRef
+                result = str(obj)
```
avdata99, line 461: Not sure on this, `unicode_safe` do not covers this type of object.

ckanext/dcat/plugins/pylons_plugin.py
```diff
@@ -32,7 +32,7 @@ def before_map(self, _map):
                          action='read_catalog')
 
             _map.connect('add dataset', '/dataset/new', controller='package', action='new')
-            _map.connect('dataset_read', '/dataset/{_id}',
+            _map.connect('dataset.read', '/dataset/{_id}',
```
amercader, line 35: I'm not sure Pylons will understand `dataset.read`, this is probably what is causing the 2.7/2.8 failures

ckanext/dcat/profiles.py
```diff
@@ -1443,7 +1442,7 @@ def _basic_fields_graph(self, dataset_ref, dataset_dict):
         self._add_date_triples_from_dict(dataset_dict, dataset_ref, items)
 
         # Dataset URL
-        dataset_url = url_for('dataset_read',
+        dataset_url = url_for('dataset.read',
```
amercader, line 1445: We might need to check the CKAN version and use `dataset_read` on CKAN < 2.9 here

ckanext/dcat/profiles.py
avdata99: There is no specific problem. Most PRs are making this change and I assume we want to leave `six.text_type`
I will back to `six` and continue with this PR @amercader 
I just moved to draft

amercader:
@avdata99 Only one minor thing to get the <2.9 tests passing:

```diff
diff --git a/ckanext/dcat/urls.py b/ckanext/dcat/urls.py
index cd209e7..7b92535 100644
--- a/ckanext/dcat/urls.py
+++ b/ckanext/dcat/urls.py
@@ -7,7 +7,7 @@ def url_for(*args, **kwargs):
     if not tk.check_ckan_version(min_version='2.9'):
 
         external = kwargs.pop('_external', False)
-        if external is not None:
+        if external is not None and 'qualified' not in kwargs:
             kwargs['qualified'] = external
 
         if len(args) and args[0] == 'dcat.read_dataset':
```

Also could you replace all usage of `qualified=True` in `url_for` with `_external=True`? Thanks!

.github/workflows/test.yml
```diff
@@ -44,6 +44,8 @@ jobs:
       CKAN_REDIS_URL: redis://redis:6379/1
 
     steps:
+    - name: adding github workspace as safe directory
+      run: git config --global --add safe.directory $GITHUB_WORKSPACE
```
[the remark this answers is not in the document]
avdata99: Because of [this](https://github.com/actions/checkout/issues/766) @amercader 
A recent change in GitHub requires this.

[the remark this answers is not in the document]
Zharktas: The checkout version could be updated instead, they claim that its fixed. At least there should be comment why its there.

ckanext/dcat/profiles.py
avdata99: Fixed with the new `urls.py` file

avdata99:
All suggestions applied @amercader 
The `qualified=True` to `_external=True` solved the problem

amercader:
Thanks @avdata99!

.github/workflows/test.yml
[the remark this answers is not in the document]
avdata99: Updated [here](https://github.com/avdata99/ckanext-dcat/commit/cb7e2091de95829e3f027df6df5d02fe7b9bb637)

ckanext/dcat/urls.py
```diff
@@ -0,0 +1,23 @@
...
+
+def url_for(*args, **kwargs):
+
+    if not tk.check_ckan_version(min_version='2.9'):
+
+        external = kwargs.pop('_external', False)
+        if external is not None and 'qualified' not in kwargs:
+            kwargs['qualified'] = external
+
+        if len(args) and args[0] == 'dcat.read_dataset':
```
ThrawnCA, line 14: Might be worth extracting `args[0]` to a local variable.

avdata99: This is already merged @ThrawnCA. Maybe you could add a new PR [here](https://github.com/ckan/ckanext-dcat/blob/master/ckanext/dcat/urls.py)

PR #210, opened 2022-04-25
Use unicode_safe validator in schema

kovalch:
Make ckantoolkit version compatible with the ckanext-harvest,
which requires ckantoolkit>=0.0.7 that have a unicode_safe function.
From ckanext-harvest:
CKAN core no longer supports builtin functions as validators.
Requires ckantoolkit>=0.0.7 for CKAN 2.8.10 support

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 rdflib==4.2.1
 rdflib-jsonld==0.4.0
 geomet>=0.2.0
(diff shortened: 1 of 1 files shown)


amercader:
Thanks @kovalch, this was an oversight on #208 

TBH, it would be good to upgrade all pinned reqs in `requirements.txt`

Issue #205, opened 2022-01-17
error in rdflib-jsonld setup command: use_2to3 is invalid.


weeix:
Hi, I'm trying to install ckanext-dcat in my [Dockerfile](https://gist.github.com/weeix/ace22ed4ca6d95deddbeb44cf6aed774). However, while installing the requirements, rdflib-jsonld 0.4.0 failed to install:

 > [extbuild 5/9] RUN pip wheel --wheel-dir=/wheels -r https://raw.githubusercontent.com/ckan/ckanext-dcat/v1.1.3/requirements.txt:                 
#8 1.905 Collecting rdflib==4.2.1
#8 2.110   Downloading rdflib-4.2.1.tar.gz (889 kB)
#8 31.44 Collecting rdflib-jsonld==0.4.0
#8 31.48   Downloading rdflib-jsonld-0.4.0.tar.gz (53 kB)
#8 33.27     ERROR: Command errored out with exit status 1:
#8 33.27      command: /usr/bin/python -c 'import io, os, sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-wheel-syjuscl3/rdflib-jsonld_7bd4471f587a4faea0e6115451d7d5e1/setup.py'"'"'; __file__='"'"'/tmp/pip-wheel-syjuscl3/rdflib-jsonld_7bd4471f587a4faea0e6115451d7d5e1/setup.py'"'"';f = getattr(tokenize, '"'"'open'"'"', open)(__file__) if os.path.exists(__file__) else io.StringIO('"'"'from setuptools import setup; setup()'"'"');code = f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-pip-egg-info-96_mnoev
#8 33.27          cwd: /tmp/pip-wheel-syjuscl3/rdflib-jsonld_7bd4471f587a4faea0e6115451d7d5e1/
#8 33.27     Complete output (1 lines):
#8 33.27     error in rdflib-jsonld setup command: use_2to3 is invalid.
#8 33.27     ----------------------------------------
[...]

PR #213, opened 2022-05-04
Upgrade RDFLib requirement

amercader:
The current version of RDFlib won't work on Python 3.9 so we need to upgrade to RDFlib 6, which is not compatible with Python 2. The JSON-LD requirement has been integrated into the main rdflib package.

This adapts the tests so they can run across all RDFLib / Python / CKAN version :crossed_fingers:

### DIFF
--- a/dev-requirements-py2.txt
+++ b/dev-requirements-py2.txt
@@ -1,0 +1,4 @@
+responses==0.17.0
+mock
+pytest-ckan
+pytest-cov
--- a/dev-requirements-py2.txt
+++ b/dev-requirements-py2.txt
@@ -1,4 +1,4 @@
-responses==0.17.0
+responses==0.10.15
 mock
 pytest-ckan
 pytest-cov
--- a/requirements-py2.txt
+++ b/requirements-py2.txt
@@ -1,5 +1,5 @@
-rdflib==4.2.1
-rdflib-jsonld==0.4.0
+rdflib==5.0.0
+rdflib-jsonld==0.5.0
 geomet>=0.2.0
 ckantoolkit>=0.0.7
 future>=0.18.2
--- a/requirements-py2.txt
+++ b/requirements-py2.txt
@@ -1,0 +1,6 @@
+rdflib==4.2.1
+rdflib-jsonld==0.4.0
+geomet>=0.2.0
+ckantoolkit>=0.0.7
+future>=0.18.2
+six
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -45,9 +45,16 @@ jobs:
 
     steps:
     - uses: actions/checkout@v3
-    - name: Install requirements
+    - name: Install requirements (Python 3)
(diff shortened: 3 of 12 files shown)


amercader:
I finally managed to get all versions green :scream_cat: 
@seitenbau-govdata (and others)  I'm fairly confident that the RDFlib upgrade won't break existing serializations and parsers but I'd be supergrateful if this PR could be tested in a real life instance with real metadata.

seitenbau-govdata:
@amercader Sure, we can do that. We can test it in the environments CKAN 2.8 under Python 2.7 and CKAN 2.9 under Python 3.6.

seitenbau-govdata:
@amercader Our environment is currently locked to Python 3.6. Unfortunately RDFlib 6 requires Python 3.7+, which means we cannot test this at the moment.

amercader:
@seitenbau-govdata It's a tricky balance. I know that Ubuntu 18.04 might be supported for a while, but Python 3.6 is no longer supported and doesn't receive security updates any more. I'm keen on maximizing support for this extension but also using recent versions to minimize issues like #205.

If I understand correctly, users running on Python 3.6 can still use the extension if they install the `requirements-py2.txt` ones right? Obviously we need to document that on the README (and maybe rename the file to `requirements-py2-py36.txt`) but do you think this would be reasonable?

seitenbau-govdata:
@amercader True, it's really tricky. Yes, the requirements in `requirements-py2.txt` still working under Python 3.6. I think this is a solution we can go with. Adding a hint in the documentation sounds good to me. Probably renaming the requirements file to `requirements-py2-py36.txt`makes it even clearer than just documenting it on the README.

Currently we are testing the branch with the dependencies `rdflib==5.0.0` and `rdflib-jsonld==0.5.0` under CKAN 2.9 (Python 3.6) and CKAN 2.8 (Python 2.7). It wouldn't hurt to update the dependencies at the same time.

dev-requirements-py2.txt
```diff
@@ -0,0 +1,4 @@
+responses==0.17.0
```
seitenbau-govdata, line 1:
```suggestion
responses==0.10.15
```

requirements-py2.txt
```diff
@@ -0,0 +1,6 @@
+rdflib==4.2.1
```
seitenbau-govdata, line 1:
```suggestion
rdflib==5.0.0
```

requirements-py2.txt
```diff
@@ -0,0 +1,6 @@
...
+rdflib-jsonld==0.4.0
```
seitenbau-govdata, line 2:
```suggestion
rdflib-jsonld==0.5.0
```

seitenbau-govdata:
@amercader And we need a rebase to the current master and adding the `self._add_responses_solr_passthru()` to the new tests
* test_harvest_default_file_size
* test_harvest_config_file_size

seitenbau-govdata:
@amercader 👍 LGTM, thanks!

amercader:
Thanks @seitenbau-govdata ! I've incorporated all your suggested changes, I think that gives a good support across versions, I'll release a new version with this and your other recent changes.

PR #214, opened 2022-05-06
Change some fields from Literal to URIRefOrLiteral

seitenbau-govdata:
We changed a few more types in the profile from Literal to URIRefOrLiteral since they could contain an URI.

### DIFF
--- a/ckanext/dcat/profiles.py
+++ b/ckanext/dcat/profiles.py
@@ -61,19 +61,20 @@ class URIRefOrLiteral(object):
     Like CleanedURIRef, this is a factory class.
     '''
     def __new__(cls, value):
-        stripped_value = value.strip()
-        if (isinstance(value, basestring) and (stripped_value.startswith("http://")
-                                               or stripped_value.startswith("https://"))):
-            uri_obj = CleanedURIRef(value)
-            # although all invalid chars checked by rdflib should have been quoted, try to serialize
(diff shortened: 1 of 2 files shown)


seitenbau-govdata:
@amercader We have changed some properties from `Literal` to `URIRefOrLiteral`, so if the value is an URI the properties will be serialized as a resource. What Do you think about the changes?

amercader:
@seitenbau-govdata this makes sense

seitenbau-govdata:
@amercader Nice! 😄 Is it ok then if we merge the pull request ourselves?

amercader:
argh, sorry @seitenbau-govdata I thought I had pressed the button!

seitenbau-govdata:
@amercader No problem. Thanks!

toothstone:
According to [the DCAT-AP.de validator](https://www.itb.ec.europa.eu/shacl/dcat-ap.de/upload) various datasets ([example](https://opendata.leipzig.de/dataset/einwohner-jahreszahlen-kleinraumig.rdf)) in our catalog are in violation since they carry dct:identifiers which are not Literals.

I'm not an RDF expert but https://www.dublincore.org/specifications/dublin-core/dcmi-terms/terms/identifier/ also says Literal.

I was about to create a PR to basically revert this PR, but had a look at the blame which brought me here. @seitenbau-govdata could you maybe elaborate why this was changed?

PR #217, opened 2022-05-25
Increase harvester get content chunk size

seitenbau-govdata:
We have some RDF endpoints which delivers a big dump file with thousands of datasets resulting in a large file of up to 100 MB. And we have noticed that getting the whole content takes a very long time and sometime even gets interrupted. The reason is the small chunk size of 1 KB. So we like to increase the chunk size up to 512 KB.


amercader:
That sounds very reasonable

PR #218, opened 2022-06-09
Fixes assert expressions

seitenbau-govdata:
The assert expressions got broken when switching from nose
to pytest (commit d4d6f6bcef9233ee6a78e5090385dc7e8ea0e4e5)


amercader:
Thanks @seitenbau-govdata

Issue #182, opened 2020-04-21
License missing


gallexme:
Hey,

what am im doing wrong if i set the Licence as usual in the Dataset, and the Distributions dont get a License Field???/the DCAT Catalogue doesnt have any license :(

shouldnt the license of the Dataset be copied to all Distributions(Aka  Ressources in CKAN?)

amercader:
Looks like the [`graph_from_dataset()`](https://github.com/ckan/ckanext-dcat/blob/master/ckanext/dcat/profiles.py#L1055) method of the base profile doesn't try to import the dataset license to the distributions (the inverse, distribution license -> dataset license, was implemented a while ago)
This would be need to be implemented.

gallexme:
Would it be possible to have it implemented, since Licenses are Very important dont you think ? :) , and reading it from the Dataset which already has the License seems like the first logical step, 
sorry i am just slightly flabbergasted that licenses are not working in a 5 year old project.


Feature idea:
1. Read License from Dataset
2. Output it in the DCAT Formats

seitenbau-govdata:
Now there is a configuration option https://github.com/ckan/ckanext-dcat#inherit-license-from-the-dataset-as-fallback-in-distributions to enable the inheritance of the license from the dataset to the distributions.

PR #231, opened 2022-08-22
Add option to inherit dataset licenses for distributions when missing

seitenbau-govdata:
We added a config-setting `ckanext.dcat.resource.inherit.license` which allows the dataset license to be used as fallback for distributions when it's missing. Only valid URIs are used as fallback.

This solves #182.


amercader:
Sorry @seitenbau-govdata I had missed this, looks good!

PR #232, opened 2022-11-21
Support dataset auto translation

wanam:
This feature depends on https://github.com/wanam/ckanext-harvest/commit/e6eb6acab777892cab6d43ad94db4d1edc3efe1b, check https://github.com/ckan/ckanext-harvest/pull/515.

It uses Google Translate free plan through deep-translator library: https://github.com/nidhaloff/deep-translator

Translation can be triggered during harvesting with "-t/--translate language" argument, where language code is ISO 639-1 which is the alpha-2 code.

Example:
ckan --config=/etc/ckan/default/ckan.ini harvester gather-consumer -t fr


amercader:
Thanks @wanam , see https://github.com/ckan/ckanext-harvest/pull/515#issuecomment-1325146838

wanam:
Understood, thank you!

PR #233, opened 2022-12-02
Add fallback for dct:issued and dct:modified in resources

seitenbau-govdata:
We'd like to add a fallback for the serialization of the resource dates _dct:issued_ and _dct:modified_, in the same way it is implemented for the dataset. To do so we use the CKAN internal dates _created_ and _metadata_modified_ of the resource.


amercader:
This makes sense @seitenbau-govdata , can you merge the latest master branch? that should fix the lint testing issues

seitenbau-govdata:
@amercader Done. The pull request is now based on the current master. Thanks for fixing the testing issues.

PR #234, opened 2022-12-05
Release version 1.4.0

seitenbau-govdata:
* Updates changelog and setup.py for version 1.4.0


seitenbau-govdata:
As there has been some changes since last realease, we would like to release a new version 1.4.0. What do you think @amercader ?

amercader:
Sounds great, go for it @seitenbau-govdata

PR #240, opened 2023-03-10
Locale Support

JVickery-TBS:
Featre/language support

Added translation support to the extensions. Translates:
- titles
- descriptions/notes
- orginization title
- resource titles
- resource descriptions
- keywords/tags

### DIFF
--- a/ckanext/dcat/converters.py
+++ b/ckanext/dcat/converters.py
@@ -1,4 +1,13 @@
 from past.builtins import basestring
+from ckan.plugins.toolkit import (
+    h,
+    get_action,
+    g,
+    abort,
+    _
+)
+from ckan import model
+from ckan.logic import NotFound
 import logging
 
 log = logging.getLogger(__name__)
@@ -58,13 +67,15 @@ def ckan_to_dcat(package_dict):
 
     dcat_dict = {}
 
-    dcat_dict['title'] = package_dict.get('title')
-    dcat_dict['description'] = package_dict.get('notes')
-    dcat_dict['landingPage'] = package_dict.get('url')
+    dcat_dict['title'] = h.get_translated(package_dict, 'title')
+    dcat_dict['description'] = h.get_translated(package_dict, 'notes')
+    dcat_dict['landingPage'] = h.url_for(package_dict.get('type', 'dataset') \
+                                         + '.read', id=package_dict.get('id'))
 
 
     dcat_dict['keyword'] = []
-    for tag in package_dict.get('tags', []):
+    tags = h.get_translated(package_dict, 'tags')
+    for tag in tags:
         dcat_dict['keyword'].append(tag['name'])
 
 
@@ -93,9 +104,16 @@ def ckan_to_dcat(package_dict):
 
     dcat_dict['distribution'] = []
     for resource in package_dict.get('resources', []):
+        context = {u'model': model, u'session': model.Session,
+                   u'user': g.user, u'auth_user_obj': g.userobj}
+        try:
+            res_dict = get_action(u'resource_show')(context, {'id': resource.get('id')})
+        except NotFound:
+            abort(404, _('Resource %s does not exist.' % resource.get('id')))
+
         distribution = {
-            'title': resource.get('name'),
-            'description': resource.get('description'),
+            'title': h.get_translated(res_dict, 'name'),
+            'description': h.get_translated(res_dict, 'description'),
             'format': resource.get('format'),
             'byteSize': resource.get('size'),
             # TODO: downloadURL or accessURL depending on resource type?
--- a/ckanext/dcat/profiles.py
+++ b/ckanext/dcat/profiles.py
@@ -1241,7 +1241,7 @@ class EuropeanDCATAPProfile(RDFProfile):
         # Tags
         tags = toolkit.h.get_translated(dataset_dict, 'tags')
         for tag in tags:
-            g.add((dataset_ref, DCAT.keyword, Literal(tag['name'])))
+            g.add((dataset_ref, DCAT.keyword, Literal(tag)))
 
         # Dates
         items = [
--- a/ckanext/dcat/profiles.py
+++ b/ckanext/dcat/profiles.py
@@ -19,6 +19,8 @@ from geomet import wkt, InvalidGeoJSONException
 
 from ckan.model.license import LicenseRegister
 from ckan.plugins import toolkit
+from ckan import model
+from ckan.logic import NotFound
 from ckan.lib.munge import munge_tag
 from ckanext.dcat.urls import url_for
 from ckanext.dcat.utils import resource_uri, publisher_uri_organization_fallback, DCAT_EXPOSE_SUBCATALOGS, DCAT_CLEAN_TAGS
@@ -157,7 +159,7 @@ class RDFProfile(object):
         '''
         Returns all DCAT keywords on a particular dataset
         '''
-        keywords = self._object_value_list(dataset_ref, DCAT.keyword) or []
+        keywords = toolkit.h.get_translated(dataset_ref, 'tags')
         # Split keywords with commas
         keywords_with_commas = [k for k in keywords if ',' in k]
         for keyword in keywords_with_commas:
@@ -991,14 +993,20 @@ class EuropeanDCATAPProfile(RDFProfile):
 
         # Basic fields
         for key, predicate in (
-                ('title', DCT.title),
-                ('notes', DCT.description),
-                ('url', DCAT.landingPage),
                 ('version', OWL.versionInfo),
                 ):
             value = self._object_value(dataset_ref, predicate)
             if value:
                 dataset_dict[key] = value
+
+        # Fluent URL workaround
+        url = url_for(dataset_dict.get('type', 'dataset') \
+                      + '.read', id=dataset_dict.get('id'), _external=True)
+        dataset_dict['url'] = url
+
+        # translation workaround
+        dataset_dict['title'] = toolkit.h.get_translated(dataset_dict, 'title')
+        dataset_dict['notes'] = toolkit.h.get_translated(dataset_dict, 'notes')
 
         if not dataset_dict.get('version'):
             # adms:version was supported on the first version of the DCAT-AP
@@ -1209,9 +1217,6 @@ class EuropeanDCATAPProfile(RDFProfile):
 
         # Basic fields
         items = [
-            ('title', DCT.title, None, Literal),
-            ('notes', DCT.description, None, Literal),
-            ('url', DCAT.landingPage, None, URIRef),
             ('identifier', DCT.identifier, ['guid', 'id'], URIRefOrLiteral),
             ('version', OWL.versionInfo, ['dcat_version'], Literal),
             ('version_notes', ADMS.versionNotes, None, Literal),
@@ -1222,8 +1227,20 @@ class EuropeanDCATAPProfile(RDFProfile):
         ]
         self._add_triples_from_dict(dataset_dict, dataset_ref, items)
 
+        # Fluent URL workaround
+        url = url_for(dataset_dict.get('type', 'dataset') \
+                      + '.read', id=dataset_dict.get('id'), _external=True)
+        g.add((dataset_ref, DCAT.landingPage, URIRef(url)))
+
+        # translation workaround
+        title = toolkit.h.get_translated(dataset_dict, 'title')
+        notes = toolkit.h.get_translated(dataset_dict, 'notes')
+        g.add((dataset_ref, DCT.title, Literal(title)))
+        g.add((dataset_ref, DCT.description, Literal(notes)))
+
         # Tags
-        for tag in dataset_dict.get('tags', []):
+        tags = toolkit.h.get_translated(dataset_dict, 'tags')
+        for tag in tags:
             g.add((dataset_ref, DCAT.keyword, Literal(tag['name'])))
 
         # Dates
@@ -1286,10 +1303,16 @@ class EuropeanDCATAPProfile(RDFProfile):
             self._get_dataset_value(dataset_dict, 'publisher_name'),
             dataset_dict.get('organization'),
         ]):
+            context = {u'model': model, u'session': model.Session,
+                        u'user': toolkit.g.user, u'auth_user_obj': toolkit.g.userobj}
+            try:
+                org_dict = toolkit.get_action(u'organization_show')(context, {u'id': dataset_dict['organization']['id']})
+            except NotFound:
+                toolkit.abort(404, toolkit._('Organization not found'))
 
-            publisher_uri = self._get_dataset_value(dataset_dict, 'publisher_uri')
+            publisher_uri = toolkit.h.url_for('organization.read', id=org_dict.get('id'))
             publisher_uri_fallback = publisher_uri_organization_fallback(dataset_dict)
-            publisher_name = self._get_dataset_value(dataset_dict, 'publisher_name')
+            publisher_name = toolkit.h.get_translated(org_dict, 'title')
             if publisher_uri:
                 publisher_details = CleanedURIRef(publisher_uri)
             elif not publisher_name and publisher_uri_fallback:
@@ -1366,8 +1389,6 @@ class EuropeanDCATAPProfile(RDFProfile):
 
             #  Simple values
             items = [
-                ('name', DCT.title, None, Literal),
-                ('description', DCT.description, None, Literal),
                 ('status', ADMS.status, None, URIRefOrLiteral),
                 ('rights', DCT.rights, None, URIRefOrLiteral),
                 ('license', DCT.license, None, URIRefOrLiteral),
@@ -1376,6 +1397,19 @@ class EuropeanDCATAPProfile(RDFProfile):
             ]
 
             self._add_triples_from_dict(resource_dict, distribution, items)
+
+            # translation workaround
+            context = {u'model': model, u'session': model.Session,
+                       u'user': toolkit.g.user, u'auth_user_obj': toolkit.g.userobj}
+            try:
+                res_dict = toolkit.get_action(u'resource_show')(context, {'id': resource_dict.get('id')})
+            except NotFound:
+                toolkit.abort(404, toolkit._('Resource %s does not exist.' % resource_dict.get('id')))
+
+            name = toolkit.h.get_translated(res_dict, 'name')
+            description = toolkit.h.get_translated(res_dict, 'description') or ''
+            g.add((distribution, DCT.title, Literal(name)))
+            g.add((distribution, DCT.description, Literal(description)))
 
             #  Lists
             items = [


smotornyuk:
@amercader, can you take a look at this PR? It's a part of PR series from @JVickery-TBS with translation fixes

ckanext/dcat/converters.py
```diff
@@ -58,13 +67,15 @@ def ckan_to_dcat(package_dict):
 
     dcat_dict = {}
 
-    dcat_dict['title'] = package_dict.get('title')
-    dcat_dict['description'] = package_dict.get('notes')
-    dcat_dict['landingPage'] = package_dict.get('url')
+    dcat_dict['title'] = h.get_translated(package_dict, 'title')
+    dcat_dict['description'] = h.get_translated(package_dict, 'notes')
+    dcat_dict['landingPage'] = h.url_for(package_dict.get('type', 'dataset') \
```
amercader, line 72: The core `url` and the dcat `landingPage` fields of the dataset metadata is not necessarily meant to be the CKAN URL for the dataset page, they can point to another source URL so I would not enforce this here

ckanext/dcat/converters.py
```diff
@@ -1,4 +1,13 @@
 from past.builtins import basestring
+from ckan.plugins.toolkit import (
+    h,
+    get_action,
+    g,
+    abort,
+    _
+)
+from ckan import model
+from ckan.logic import NotFound
```
amercader, line 10: Please use `toolkit.ObjectNotFound` instead

ckanext/dcat/converters.py
```diff
@@ -93,9 +104,16 @@ def ckan_to_dcat(package_dict):
 
     dcat_dict['distribution'] = []
     for resource in package_dict.get('resources', []):
+        context = {u'model': model, u'session': model.Session,
+                   u'user': g.user, u'auth_user_obj': g.userobj}
```
amercader, line 108: Just passing `user` should be enough

ckanext/dcat/converters.py
```diff
@@ -93,9 +104,16 @@ def ckan_to_dcat(package_dict):
...
+        try:
+            res_dict = get_action(u'resource_show')(context, {'id': resource.get('id')})
```
amercader, line 110: Why do we need to call `resource_show`  here?

ckanext/dcat/profiles.py
```diff
@@ -991,15 +993,21 @@ def parse_dataset(self, dataset_dict, dataset_ref):
...
                 ('version', OWL.versionInfo),
                 ):
             value = self._object_value(dataset_ref, predicate)
             if value:
                 dataset_dict[key] = value
 
+        # Fluent URL workaround
+        url = url_for(dataset_dict.get('type', 'dataset') \
+                      + '.read', id=dataset_dict.get('id'), _external=True)
+        dataset_dict['url'] = url
```
amercader, line 1005: See comment above for the `url` field

ckanext/dcat/profiles.py
```diff
@@ -991,15 +993,21 @@ def parse_dataset(self, dataset_dict, dataset_ref):
 
         # Basic fields
         for key, predicate in (
-                ('title', DCT.title),
-                ('notes', DCT.description),
-                ('url', DCAT.landingPage),
```
amercader, line 996: The `parse_dataset()` method essentially maps the RDF graph from the remote semantic serialization to a CKAN dataset dict. If we remove these lines we are not extracting the remote values anymore. If we just do 

     dataset_dict['title'] = toolkit.h.get_translated(dataset_dict, 'title')

I don't see how we can actually obtain the remote value for `title`. And besides, IIUC  the only thing `h.get_translated()` does is looking for an existing `title_translated[lang]` field in the dataset dict, which at this point won't exist.

Without knowing what you are trying to achieve in detail, I guess if the final goal is to create a dataset dict with fluent-like metadata fields, we would need to
1. Parse whatever way DCAT/RDF has of defining multilingual metadata
2. Use that information to create the relevant `*_translated` fields

This would only work if the user is using ckanext-fluent, so perhaps that should be an opt-in behaviour

ckanext/dcat/profiles.py
```diff
@@ -1209,9 +1217,6 @@ def graph_from_dataset(self, dataset_dict, dataset_ref):
 
         # Basic fields
         items = [
-            ('title', DCT.title, None, Literal),
```
amercader, line 1212: In the `grap_from_dataset()` case we want to do the opposite, translate the `_translated` fields into whatever triples are needed to express multilingual metadata. At the very least I would keep these lines to provide a default fallback for the serialization of these fields, and then if necessary, handle the `_translated` ones

ckanext/dcat/profiles.py
```diff
@@ -1222,9 +1227,21 @@ def graph_from_dataset(self, dataset_dict, dataset_ref):
         ]
         self._add_triples_from_dict(dataset_dict, dataset_ref, items)
 
+        # Fluent URL workaround
+        url = url_for(dataset_dict.get('type', 'dataset') \
```
amercader, line 1231: See note about `landingPage` field

ckanext/dcat/converters.py
JVickery-TBS: @amercader so the `package_dict.get('url')` would return a possible relative url? But it would be possible for some ckan users to have a package's url to an external url?

Would it be possible to check if the `package_dict.get('url')` is a relative address, and then do the `url_for` as the `url_for` here is to include any language subdomains or sub-directories in use.

ckanext/dcat/converters.py
JVickery-TBS: N/A not needed anymore.

ckanext/dcat/converters.py
JVickery-TBS: N/A not needed anymore.

ckanext/dcat/converters.py
JVickery-TBS: Sorry I was bringing code over from a different fork, and that was on an old ckan version. Tested without the `resource_show` and it works as expected.

ckanext/dcat/profiles.py
JVickery-TBS: @amercader okay, so will I need to even do anything in the `profiles.py` even? Is everything related to what I need to do this fluent stuff just in the `converters.py`?

JVickery-TBS:
@amercader okay I understand this extension a bit better from your comments thanks!

So basically, we do not want to really mess with the Profiles at all because they are just the mappings of all of the different meta formats.

So I found that the logic method `dcat_dataset_show` here is what is used. And to kind of go off more of your comments, I leaned into making these really just "Fluent Compatibility" because that is what it is really.

So I now check `plugin_loaded('fluent')`.

I have reverted everything from `profiles.py` and the code in the `dcat_dataset_show` seems to handle everything well as that is where we can modify the data dict values returned from `package_show` before the dict is passed through the profiles/serialization.

I realize that it is not the greatest level/place to put the code, but it is clean and easy.

The only thing that I would really like to do still is figure out the `landingPage` for the correct language URLs. Because this is an example of the current code for a url such as http://127.0.0.1:5009/fr/dataset/d9865e61-227c-4298-a1d1-620e6669b097.xml:
```
<rdf:RDF>
  <dcat:Dataset rdf:about="http://127.0.0.1:5009/dataset/d9865e61-227c-4298-a1d1-620e6669b097">
    <dct:title>Feeds Testing FR</dct:title>
    <dct:description>Feeds Testing Description FR</dct:description>
```
So it would be nice to have the `landingPage` be http://127.0.0.1:5009/fr/dataset/d9865e61-227c-4298-a1d1-620e6669b097

Let me know if you think there is any way to do that?

ckanext/dcat/profiles.py
JVickery-TBS: N/A `profiles.py` reverted.

ckanext/dcat/profiles.py
JVickery-TBS: N/A `profiles.py` reverted.

ckanext/dcat/profiles.py
JVickery-TBS: N/A `profiles.py` reverted.

ckanext/dcat/profiles.py
JVickery-TBS: N/A `profiles.py` reverted.

ckanext/dcat/converters.py
JVickery-TBS: Resolving for now, new comments regarding the `landingPage` below...might not be able to accomplish my desires though.

JVickery-TBS:
@amercader Okay I figured it out now! Firstly, I added the fluent compatibility to the catalog view/blueprint as well now.

As for the landingPage/URL stuff, I figured that out. So it will not override any set `url` field values or extra field values. As an example:
* I have a dataset with any empty `url` field;
* The extension calls `dataset_uri` (https://github.com/ckan/ckanext-dcat/blob/master/ckanext/dcat/utils.py#L178)
* Because of the empty `url` fields, it will generate one, calling `catalog_uri` method (https://github.com/ckan/ckanext-dcat/blob/master/ckanext/dcat/utils.py#L141)

```
 The value will be the first found of:
        1. The value of the `uri` field
        2. The value of an extra with key `uri`
        3. `catalog_uri()` + '/dataset/' + `id` field
```

It is now in the `catalog_uri` where I check if fluent is loaded, and that the user is within a request context to replace the `{{LANG}}` tag in the config options `ckan.site_url` or `ckanext.dcat.base_uri` with the current language.

seitenbau-govdata:
Maybe similar to https://github.com/ckan/ckanext-dcat/pull/124

amercader:
Thanks @JVickery-TBS . Let's step back for a second and see what means for ckanext-dcat to have multi-language support.
We will assume that for a field to be multilingual it needs to use the ckanext-fluent convention:

```
dataset_dict = {
    "title": {
        "en": "Some title in English",
        "ca": "Un títol en català",
        "es": "Un título en castellano",
    },
    "notes": {
        "en": "A description in English",
        "ca": "Una descripció en català",
        "es": "Una descripción en castellano",
    },
    "resources": [],
    "maintainer": "xx",

}
```
Multilingual support means:

1. Supporting importing fields from the RDF graph to the fluent format above (which is the goal of #124 by @stefina)
2. Using the fluent fields to serialize multilingual RDF files (which is the goal of this PR)

Both don't need to be done at the same time, so it's fine to focus on serialization for now.
You current approach is to modify the values of multilingual fields in the serialized RDF file (the RDF/XML, jsonld, ttl...) to match the language that the current web user is using. So if they are visiting https://someckan.org/en they get:

```xml
<rdf:RDF>
  <dcat:Dataset rdf:about="https://someckan.org/dataset/d9865e61-227c-4298-a1d1-620e6669b097">
    <dct:title>Some title in English</dct:title>
    <dct:description>A description in English</dct:description>
```

And if they are visiting https://someckan.org/ca (or `ckan.locale_default = ca`) they will get:

```xml
<rdf:RDF>
  <dcat:Dataset rdf:about="https://someckan.org/dataset/d9865e61-227c-4298-a1d1-620e6669b097">
    <dct:title>Un títol en català</dct:title>
    <dct:description>Una descripció en català</dct:description>
```

This is very limited in that we are not representing all languages, and the serialization does not provide information on which is the actual language used to display the values are provided. Besides, this will only work in the context of a web request, not when used in the CLI, or as module elsewhere.

The correct and more interoperable approach is to always provide all available languages, and use language codes:

```xml
<rdf:RDF>
  <dcat:Dataset rdf:about="https://someckan.org/dataset/d9865e61-227c-4298-a1d1-620e6669b097">
    <dct:title xml:lang="en">Some title in English</dct:title>
    <dct:title xml:lang="ca">Un títol en català</dct:title>
    <dct:title xml:lang="es">Un título en castellano</dct:title>
    <dct:description xml:lang="en">A description in English</dct:description>
    <dct:description xml:lang="ca">Una descripció en català</dct:description>
    <dct:description xml:lang="es">Una descripción en castellano</dct:description>
```

For this I'm afraid we need to go low level, at the profiles level. But the good news is that by changing it there we will automatically get multilingual serializations regardless of how these are created (API, RDF endpoint, CLI, etc).

Below is a quick patch I tried to add support for multilingual `title` and `notes` fields. Hopefully it's easy to expand to other fields. @JVickery-TBS if you could give it a go I believe that could be a good path forward.

@seitenbau-govdata I'd love to get another pair of eyes on the modified `_add_triple_from_dict()` logic. I think the assumptions I made are fair but maybe we need to consider other combinations of parameters

```diff
diff --git a/ckanext/dcat/profiles.py b/ckanext/dcat/profiles.py
index 9b066ef..f91c79e 100644
--- a/ckanext/dcat/profiles.py
+++ b/ckanext/dcat/profiles.py
@@ -727,14 +727,16 @@ class RDFProfile(object):
 
     def _add_triples_from_dict(self, _dict, subject, items,
                                list_value=False,
-                               date_value=False):
+                               date_value=False,
+                               multilingual=False):
         for item in items:
             key, predicate, fallbacks, _type = item
             self._add_triple_from_dict(_dict, subject, predicate, key,
                                        fallbacks=fallbacks,
                                        list_value=list_value,
                                        date_value=date_value,
-                                       _type=_type)
+                                       _type=_type,
+                                       multilingual=multilingual)
 
     def _add_triple_from_dict(self, _dict, subject, predicate, key,
                               fallbacks=None,
@@ -742,7 +744,8 @@ class RDFProfile(object):
                               date_value=False,
                               _type=Literal,
                               _datatype=None,
-                              value_modifier=None):
+                              value_modifier=None,
+                              multilingual=False):
         '''
         Adds a new triple to the graph with the provided parameters
 
@@ -776,6 +779,11 @@ class RDFProfile(object):
             self._add_date_triple(subject, predicate, value, _type)
         elif value:
             # Normal text value
+            if multilingual and isinstance(value, dict):
+                # We assume that all multilingual field values are Literals
+                for lang, translated_value in value.items():
+                    object = Literal(translated_value, lang=lang)
+                    self.g.add((subject, predicate, object))
             # ensure URIRef items are preprocessed (space removal/url encoding)
             if _type == URIRef:
                 _type = CleanedURIRef
@@ -1207,10 +1215,16 @@ class EuropeanDCATAPProfile(RDFProfile):
 
         g.add((dataset_ref, RDF.type, DCAT.Dataset))
 
-        # Basic fields
+        # Multilingual fields
         items = [
             ('title', DCT.title, None, Literal),
             ('notes', DCT.description, None, Literal),
+        ]
+
+        self._add_triples_from_dict(dataset_dict, dataset_ref, items, multilingual=True)
+
+        # Basic fields
+        items = [
             ('url', DCAT.landingPage, None, URIRef),
             ('identifier', DCT.identifier, ['guid', 'id'], URIRefOrLiteral),
             ('version', OWL.versionInfo, ['dcat_version'], Literal),
```

JVickery-TBS:
@amercader Okay here we go again hahaha.

I have done your above implementation for `multilingual` in the `_add_triples_from_dict` method. And it seems to be working nicely.

There were a couple places in which I had to do some strange-ish things:
- tags: because the tags are a dict, I had to loop through those in a specific way;
- organization: the publisher fallback onto the organization name, the org/group translated fields are not included in the dataset dict, so I had to call an org show here.

I also removed the language from the url for the `rdf:about=` as that seemed incorrect?

JVickery-TBS:
@amercader Just added in the fallback for the field keys. Realized that if we put in the `_translated` keys, we would be assuming that a user is translated all of these field. E.g. a user could be translating the Resource Title, but not translating the Resource Description.

So just a simple check if the `_translated` key is in the object dicts, kind of like how the core `get_translated` works.

JVickery-TBS:
@amercader Ian mentioned that I should rename the `multilingual` parameter because that is the name of the core extensions Multilingual. 

So I just renamed that param to `all_translated`.

ckanext/dcat/profiles.py
```diff
@@ -1306,9 +1328,16 @@ def graph_from_dataset(self, dataset_dict, dataset_ref):
             # If no name but an URI is available, the name literal remains empty to
             # avoid mixing organization and dataset values.
             if not publisher_name and not publisher_uri and dataset_dict.get('organization'):
-                publisher_name = dataset_dict['organization']['title']
-
-            g.add((publisher_details, FOAF.name, Literal(publisher_name)))
+                try:
+                    org_dict = toolkit.get_action(u'organization_show')({u'user': toolkit.g.user},
```
amercader, line 1332: It's a bit annoying that we need to have this call as it slows things down. We need to implement some sort of caching because the serialization generally happens on large numbers of datasets and we can't keep calling the same action over and over. Something like a class level `self._org_cache` that gets reset when the class is initialized.

```
org_id = dataset_dict["organization"]["id"]
if org_id in self._org_cache:
    org_dict = self._org_cache[org_id]
else: 
    org_dict = toolkit.get_action("organization_show")({"ignore_auth": True}, {"id": org_id})
    self._org_cache[org_id] = org_dict
```
Also note we need to use `{"ignore_auth": True}` otherwise this won't work outside a web request (eg [tests](https://github.com/ckan/ckanext-dcat/actions/runs/5321585352/jobs/9636871682))

You'll also need to tweak this tests to get it passing:

```diff
diff --git a/ckanext/dcat/tests/test_euro_dcatap_profile_serialize.py b/ckanext/dcat/tests/test_euro_dcatap_profile_serialize.py
index e0f7d3c..d9bbe47 100644
--- a/ckanext/dcat/tests/test_euro_dcatap_profile_serialize.py
+++ b/ckanext/dcat/tests/test_euro_dcatap_profile_serialize.py
@@ -398,13 +398,16 @@ class TestEuroDCATAPProfileSerializeDataset(BaseSerializeTest):
         assert self._triple(g, publisher, DCT.type, URIRef(extras['publisher_type']))
 
     def test_publisher_org(self):
+
+        org = factories.Organization()
+
         dataset = {
             'id': '4b6fe9ca-dc77-4cec-92a4-55c6624a5bd6',
             'name': 'test-dataset',
             'organization': {
-                'id': '',
-                'name': 'publisher1',
-                'title': 'Example Publisher from Org',
+                'id': org["id"],
+                'name': org["name"],
+                'title': org["title"]
             }
         }
```

amercader (commented):
Thanks for bearing with me @JVickery-TBS 

I don't like `all_translated`, what about just `translated`? 

And one final thing about the `organization_show` call. Besides that this is good to go

inderps:
Will this ever get merged or not?

ckanext/dcat/profiles.py
JVickery-TBS: @amercader SURPRISE! Back to this one! Haha, had just a little free time before our upgrade to Python3 and CKAN 2.10

Wow, this was from June eh? WELL

There is now a class variable to cache the org dicts. I renamed `all_translated` to just `translated`. AND fixed that test above

JVickery-TBS:
@inderps Hey! Sorry! Have been busy with things over here. But have just done the feedback now, so we shall see!

amercader:
Quick update here just to say that I've pulled the fluent compatibility work in the wider Scheming / DCAT 3 support effort so at some point during the next few weeks this will be looked at. I just need to think about how it will integrate with the more general scheming support but the majority of the work here should get incorporated as is. Thanks for bearing with me @JVickery-TBS

PR #241, opened 2023-03-16
Update ckan hooks for CKAN 2.10 support

seitenbau-govdata:


seitenbau-govdata:
@amercader The tests for CKAN < 2.9 and with Python 2 are now broken after merging https://github.com/ckan/ckanext-harvest/pull/520 in `ckanext-harvest`. I assume that the tests on the master are also broken. So the changes in the pull request are independent from the failing tests.

Is there a plan to update the `ckanext-dcat` according to https://github.com/ckan/ckanext-harvest/pull/520 in `ckanext-harvest`?

seitenbau-govdata:
@pdelboca Ok. We'll try to make the necessary changes in a new PR.

seitenbau-govdata:
@pdelboca and @amercader If you agree, we would like to merge the pull request as a base for https://github.com/ckan/ckanext-dcat/pull/244, because here the translation tests fail.

PR #261, opened 2023-11-30
Fix handling of duplicate pages in RDF harvester's gather_stage

### LINKED ISSUE #260: Improving Pagination Handling in RDF Harvester's gather_stage
The `gather_stage` function in the [CKAN RDF harvester](https://github.com/ckan/ckanext-dcat/blob/1109205069dd105dda27e3486898e4ca1525a808/ckanext/dcat/harvesters/rdf.py#L145) appears to have a limitation in handling incorrect pagination entities within `hydra:PagedCollection`. Specifically, when `hydra:nextPage` resources reference pages that have already been visited, the harvesting process may enter into an infinite loop, causing it to continue until the system's RAM capacity is reached and the job is terminated.

thorge:
Fixes #260

### DIFF
--- a/ckanext/dcat/harvesters/rdf.py
+++ b/ckanext/dcat/harvesters/rdf.py
@@ -146,6 +146,8 @@ class DCATRDFHarvester(DCATHarvester):
 
         log.debug('In DCATRDFHarvester gather_stage')
 
+        visited_urls = set()
+
         rdf_format = None
         if harvest_job.source.config:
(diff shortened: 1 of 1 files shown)


seitenbau-govdata:
@thorge Many thanks for the contribution. LGTM. This might be helpful in some situations with buggy implementations. If you could add at least one test case for the changed behavior it would be a pleasure to merge it. What do you think @amercader ?

PR #262, opened 2024-01-19
Add support for high value datasets

seitenbau-govdata:
This Pull request adds support for High Value Datasets as specified in https://semiceu.github.io/DCAT-AP/releases/2.2.0-hvd/

- Add `dcatap.applicableLegislation` and `dcatap.hvdCategory` to `dcat:Dataset`
- Add `dcatap.applicableLegislation` to `dcat:Distribution`

Note: Tests are currently failing until #263 is solved.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased](https://github.com/ckan/ckanext-dcat/compare/v1.5.1...HEAD)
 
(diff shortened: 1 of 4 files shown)


toothstone:
> The Tests are now working properly. Does anyone want to take a look at this PR? Otherwise we`ll merge it soon and create a new version.

I don't have the resources to run the code, but at least from reading the diff I would approve of the changes. They fit my understanding of the DCAT spec and I would have implemented it the same way.

Looking forward to see this released!

PR #264, opened 2024-01-26
[WIP] Add DCAT-AP HVD properties

toothstone:
The aim is to properly parse, store, and serialize the new DCAP-AP HVD properties for Datasets and Distributions.


toothstone:
#262 has already done it

PR #280, opened 2024-05-08
Catch Not Authorized in View

JVickery-TBS:
fix(views): catch not auth;

- Catch not authorized exception and return 403 abort.


amercader:
Thanks @JVickery-TBS !

PR #281, opened 2024-05-22
Scheming support

amercader:
This PR adds initial support for seamless integration between ckanext-dcat and ckanext-scheming, providing a custom profile that modifies the dataset dicts generated and consumed from the existing profiles so it plays well with the scheming presets defined.

## Summary of changes

* A scheming schema definition with the different DCAT properties defined (`ckanext/dcat/schemas/dcat_ap_2.1.yaml`)
* Helper functions in the `RDFProfile` class to access schema field definitions from datasets and resources
* A new `euro_dcat_ap_scheming` profile that adds support for the field serializations supported by the ckanext-scheming presets. The existing profiles (`euro_dcat_ap` and `euro_dcat_ap_2`) remain unchanged (except for some very minor backward compatible changes regarding the handling of access services in distributions/resources). This means that existing sites will keep working as currently, but maintainers can choose to enable scheming support if they choose to migrate to that approach. Upcoming DCAT 3 based profiles will be scheming based (in a new ckanext-dcat version)
* Includes "end to end" tests for DCAT -> CKAN and CKAN -> DCAT using the new schema to ensure it works as expected

## Compatibility and release plan

Extra care has been taken to not break any existing systems. Sites using the existing `euro_dcat_ap` and `euro_dcat_ap_2` profiles should not see any change in their current parsing and serialization functionalities and these profiles will never change their outputs. Sites willing to migrate to a scheming based profile can do so by adding the new `euro_dcat_ap_scheming` profile at the end of their profile chain (value of `ckanext.dcat.rdf.profiles` config option, eg `ckanext.dcat.rdf.profiles = euro_dcat_ap_2 euro_dcat_ap_scheming`), which will modify the existing profile outputs to the expected format by the scheming validators. Note that the scheming profile will only affect fields defined in the schema definition file, so sites can start migrating gradually different metadata fields.

This compatibility profile will be released in the next ckanext-dcat version (1.8.0). The upcoming DCAT v3 based profiles for DCAT-AP 3 and DCAT-US 3 will be scheming based and will incorporate the mapping changes described below.


## Mapping changes

The main changes between the old processors (parsers and serializers) and the new scheming-based ones are:

### Root level fields

Custom DCAT fields that didn't link directly to standard CKAN fields were stored as extras (see all the ones marked `extra:` [here](https://github.com/ckan/ckanext-dcat?tab=readme-ov-file#rdf-dcat-to-ckan-dataset-mapping)). So the DCAT `version_notes` field would be stored as:

```json
{
    "name": "test_dataset_dcat",
    "title": "Test dataset DCAT",
    "extras": [
         {"key": "version_notes", "value": "Some version notes"}
    ]
}
```
In the scheming-based profile, if the field is defined in the scheming schema, it will get stored as a root level field, like all custom dataset properties:


```json
{
    "name": "test_dataset_dcat",
    "title": "Test dataset DCAT",
    "version_notes": "Some version notes"
}
```

### List fields

The old profiles stored lists as JSON strings:

```json
{
    "name": "test_dataset_dcat",
    "title": "Test dataset DCAT",
    "extras": [
         {"key": "conforms_to", "value":"[\"Standard 1\", \"Standard 2\"]"}
    ],
    "resources": [
        {
             "name": "Some resource",
             "documentation": "[\"http://dataset.info.org/distribution1/doc1\", \"http://dataset.info.org/distribution1/doc2\"]"
        }
    ]
}
```

By using the [`multiple_text`](https://github.com/ckan/ckanext-scheming?tab=readme-ov-file#preset) preset, lists are now automatically handled:


```json
{
    "name": "test_dataset_dcat",
    "title": "Test dataset DCAT",
    "conforms_to": [
         "Standard 1", 
         "Standard 2"
    ],
    "resources": [
        {
             "name": "Some resource",
             "documentation": [
                 "http://dataset.info.org/distribution1/doc1", 
                 "http://dataset.info.org/distribution1/doc2"
             ]
        }
    ]
}

```

The form snippets UI allows to provide multiple values:

![Screenshot 2024-05-22 at 10-15-58 Dataset - CKAN](https://github.com/ckan/ckanext-dcat/assets/200230/6192c995-661c-4dbb-b79a-2df90bbd7382)

### Repeating subfields

Mapping complex entities like `dcat:contactPoint` or `dct:publisher` was very limited, storing a subset of properties of just one linked entity as prefixed extras:

```json
{
    "name": "test_dataset_dcat",
    "title": "Test dataset DCAT",
    "extras": [
        {"key":"contact_name","value":"PointofContact"},
        {"key":"contact_email","value":"contact@some.org"}
    ],
}
```

By using the [`repeating_subfields`](https://github.com/ckan/ckanext-scheming?tab=readme-ov-file#preset) preset we can consume and present these as proper objects, and store multiple entities for those properties that have 0..n cardinality (see comment in "Issues"):

```json
{
    "name": "test_dataset_dcat",
    "title": "Test dataset DCAT",
    "contact": [
        {
            "name": "Point of Contact 1",
            "email": "contact1@some.org"
        },
        {
            "name": "Point of Contact 2",
            "email": "contact2@some.org"
        },
    ]
}
```

Repeating subfields are also supported in resources/distributions. In this case complex objects like `dcat:accessService` were stored as JSON strings:

```json
{
    "name": "test_dataset_dcat",
    "title": "Test dataset DCAT",
    "resources": [
        {
             "name": "Some resource",
             "access_services": "[{\"availability\": \"http://publications.europa.eu/resource/authority/planned-availability/AVAILABLE\", \"title\": \"Sparql-end Point\", \"endpoint_description\": \"SPARQL url description\", \"license\": \"http://publications.europa.eu/resource/authority/licence/COM_REUSE\", \"access_rights\": \"http://publications.europa.eu/resource/authority/access-right/PUBLIC\", \"description\": \"This SPARQL end point allow to directly query the EU Whoiswho content (organization / membership / person)\", \"endpoint_url\": [\"http://publications.europa.eu/webapi/rdf/sparql\"], \"uri\": \"\", \"access_service_ref\": \"N2ff5798aac56447e89438cc838512d26\"}]"
        }
    ]
}
```
They now appear as proper objects:

```json
{
    "name": "test_dataset_dcat",
    "title": "Test dataset DCAT",
    "resources": [
        {
             "name": "Some resource",
             "access_services": [                                                                                                                                                                                                                                                 
                    {                                                                                                                                                                                                                                                                
                        "availability": "http://publications.europa.eu/resource/authority/planned-availability/AVAILABLE",                                                                                                                                                           
                        "title": "Sparql-end Point",                                                                                                                                                                                                                                 
                        "endpoint_description": "SPARQL url description",                                                                                                                                                                                                            
                        "license": "http://publications.europa.eu/resource/authority/licence/COM_REUSE",                                                                                                                                                                             
                        "access_rights": "http://publications.europa.eu/resource/authority/access-right/PUBLIC",                                                                                                                                                                     
                        "description": "This SPARQL end point allow to directly query the EU Whoiswho content (organization / membership / person)",                                                                                                                                 
                        "endpoint_url": [                                                                                                                                                                                                                                            
                            "http://publications.europa.eu/webapi/rdf/sparql"                                                                                                                                                                                                        
                        ],                                                                                                                                                                                                                                                           
                        "uri": "",                                                                                                                                                                                                                                                   
                    }                                                                                                                                                                                                                                                                
                ]
        }
    ]
}
```

Again, these can be easily managed via the UI thanks to the scheming form snippets:

![Screenshot 2024-05-22 at 10-56-35 Dataset - CKAN](https://github.com/ckan/ckanext-dcat/assets/200230/47c407fd-1383-4a5a-b2dd-e6b0969f5352)

### Issues 
* For complex objects like `dct:publisher` that have 0..1 cardinality, I don't think CKAN supports "non-repeating" subfields so it makes sense to use the `repeating_subfields` one for now and create a new one in the future.
* Scheming has presets for `date` and `datetime` with nice UI form snippets so it's tempting to use them for properties like `issued` and modified, but these support other formats like `xsd:gYear` or `xsd:gYearMonth` which will fail with these presets so we can consider creating a new one that extends the existing ones to support these formats


EricSoroos:
@amercader I've been working in DCAT this week, including adding spec compliant HVD 2.2.0 output and scheming portions to the current 1.7 version. (somewhat split across dcat and our schema extension at the moment).

A couple of things have come up for making the output compliant with the HVD shaql files:

1) There are some items that need to be typed, e.g. licenses. This is a first cut, and I want to refactor this into the add_triples... methods: https://github.com/derilinx/ckanext-dcat/blob/dcat-hvd-2.2.0/ckanext/dcat/profiles.py#L914

```
 def _add_with_class(self, dataset_dict, dataset_ref, key, predicate, _type, _class, list_value=False):
        value = self._get_dataset_value(dataset_dict, key)

        def _add(v):
            ref = _type(v)
            self.g.add((ref, RDF.type, _class))
            self.g.add((dataset_ref, predicate, ref))

        if value:
            if list_value:
                for v in self._read_list_value(value):
                    _add(v)
            else:
                _add(value)
...
            self._add_with_class(resource_dict, distribution, 'license', DCT.license, URIRefOrLiteral, DCT.LicenseDocument)
```

gives us something like this:

```
...
<http://www.opendefinition.org/licenses/cc-by> a dct:LicenseDocument .

<http://data.europa.eu/eli/reg_impl/2023/138/oj> a <http://data.europa.eu/eli/ontology#LegalResource> .

<https://test.staging.derilinx.com/dataset/242e33cf-a097-4f59-94f3-25fcddeffaec/resource/52dcb446-d1f1-40d2-a515-bd708a57b9c6> a dcat:Distribution ;
    dcatap:applicableLegislation <http://data.europa.eu/eli/reg_impl/2023/138/oj> ;
    dct:format "HTML" ;
    dct:issued "2024-05-20T16:12:07"^^xsd:dateTime ;
    dct:license <http://www.opendefinition.org/licenses/cc-by> ;
    dct:modified "2024-05-20T17:00:51"^^xsd:dateTime ;
    dct:title "Test" ;
    dcat:accessURL <https://test.staging.derilinx.com/> .
```

2) Codelists are important, e.g., the HVD Category needs to be from this list: https://op.europa.eu/en/web/eu-vocabularies/dataset/-/resource?uri=http://publications.europa.eu/resource/dataset/high-value-dataset-category  (which when it's not being slammed, has an RDF file with a skos:Concept and entries, each with a prefLabel from each official EU language.)

amercader:
@EricSoroos thanks for the feedback:

1. HVD 2.20 support sounds amazing. Is this developed in a separate profile built on top of `euro_dcat_ap_2`? If so would be great to have it upstream. I see it as not directly related to this PR, once support for DCAT-AP 2.1 is ready we can create a separate profile and schema for HDV 2.2.0
2. More generally regarding SHACL validation I've been thinking about integrating it as part of the test suite or even as a command that site maintainers can run as a way to "certify" support for the different DCAT specs
3. Types: absolutely. I literally was thinking about this today with relation to the different types that dates can have according to the spec (see "Issues" in the description), but if it's a requirement of the SHAQL validation even more so. I like the approach of extending the `_add_triples_...` utility functions 
4. Codelists: yes, that's definitely on the list for the DCAT v3 profiles, as there are controlled vocabularies used but of course it also makes sense if needed for HDV. We could explore using `choices` or more likely [`choices_helper`](https://github.com/ckan/ckanext-scheming?tab=readme-ov-file#choices_helper) presets for required or recommended fields, with CLI commands to import them into the main or datastore database, plus a form snippet that shows the options (or autocompletes them if there are a lot of them)

Great to see you are working on this same area. If you have any feedback on the general approach followed for scheming support it would be great to hear it.

amercader:
@seitenbau-govdata @bellisk would love to know your take on this, and see if this approach would play well with how you are using ckanext-dcat

ckanext/dcat/plugins/__init__.py
```diff
@@ -132,6 +135,30 @@ def set_titles(object_dict):
...
+        except KeyError:
+            pass
+
+        if schema:
+            for field in schema['dataset_fields']:
+                if field['field_name'] in dataset_dict and 'repeating_subfields' in field:
+                    for index, item in enumerate(dataset_dict[field['field_name']]):
+                        for key in item:
+                            # Index a flattened version
+                            new_key = f'{field["field_name"]}_{index}_{key}'
```
wardi, line 155: IIUC this would let us ask solr for things like 'datasets with the 6th name field in contacts equal to "frank"', but not 'datasets containing any contact named "frank"'. If we the same keys from all subfields into a single field like
```
new_key = f'{field["field_name"]}__{key}'
```
then we could, right? We're not doing dynamic solr schemas so we'd have to combine everything as text fields but it should work for text searches.

Also nothing here is specific to dcat so shouldn't it be a scheming PR?

ckanext/dcat/processors.py
```diff
@@ -56,6 +56,8 @@ def __init__(self, profiles=None, compatibility_mode=False):
             raise RDFProfileException(
                 'No suitable RDF profiles could be loaded')
 
+        self.dataset_schema = dataset_schema
```
wardi, line 59: dataset_type?

ckanext/dcat/schemas/dcat_ap_2.1.yaml
```diff
@@ -0,0 +1,141 @@
...
+
+    - field_name: uri
+      label: URI
+
+    - field_name: title
+      label: Title
+
+    - field_name: endpoint_url
+      label: Endpoint URL
+      preset: multiple_text
```
wardi, line 141: WDYT about tighter validation on this schema, e.g. BCP47 values in language, valid emails, URIs and URLs?

ckanext/dcat/plugins/__init__.py
amercader: Right, given this:

Screenshot 2024-05-23 at 13-26-38 Test dcat 1 - Dataset - CKAN

With your suggestion you need to run a query like `q=contact__name:*Racoon*` to get a hit, with the original it would be impossible to get this hit without knowing the index, so that's obviously better. 
But I think users would expect to find these results in a free text search as well. For this, the subfields need to be indexed in an `extras_*` field, as these are copied to the [catch-all `text`](https://github.com/ckan/ckan/blob/5334babe3888a239dea15a17c8af8e46d1c3f346/ckan/config/solr/schema.xml#L186) field. So this would allow just to do `q=Racoon` and get a result back, so maybe indexing 

```
new_key = f'extras_{field["field_name"]}__{key}'
```
with the combined key values is the better approach.

> Also nothing here is specific to dcat so shouldn't it be a scheming PR?

Sure, I was just getting things working in this extension. Do you want to replace the logic in [https://github.com/ckan/ckanext-scheming/blob/9a33dcd330211468ea0bed85391277e2e596f388/ckanext/scheming/plugins.py#L477](https://github.com/ckan/ckanext-scheming/blob/9a33dcd330211468ea0bed85391277e2e596f388/ckanext/scheming/plugins.py#L477) with this one or create a separate plugin? I'll send a PR

ckanext/dcat/schemas/dcat_ap_2.1.yaml
amercader: I think it's a great idea, but I want to do it in a second stage once all the fields are defined in the schema and the general approach validated. I'll start to compile a list of possible validations, these are all great candidates

EricSoroos:
A little more thinking on the relationship between DCAT-AP base and the extension profiles (e.g. HVD, Geo, etc) .

I think that it would definitely make sense to have the individual profiles have either pluggable schema sections or diffs/inheritance against the core schema. E.g., Site A needs HVD, Site B needs HVD + Geo.  We're using our schema_field_groups for this, so there's an HVD tab in the dataset view. 

image


At the graph generation level, I don't know if there's a clean way to do this in an inherited manner. Right now, the EUDCAT2 is a combination of v1 + base + HVD + Geo.   There's no issue adding the additional profiles here if the data doesn't support it.

Maybe a better way to do this would be composition rather than inheritance. E.g., have the profile configure a set of [ckan object]_to_graph methods, and those additional profiles would only be responsible for those items that aren't part of the base. As it is, it feels like the profile inheritance is quite chunky for adding a few fields.

wardi:
I deliberately avoided implementing inheritance in scheming. Instead dcat fields can be defined as presets (including the field ids) and specific schemas then populate fields from the presets like
```
- preset: dcat_dataset_contact
- preset: dcat_dataset_publisher
...
```
This gives complete control from the specific schemas without needing to resolve inheritance issues.

EricSoroos:
Fair enough -- It's not too hard to make sure that you have all of the chunks required for a specific profile, though for HVD they're scattered over both the dataset and resource.

amercader:
Thanks @EricSoroos I need to think a bit more about how best to support extensions but I don't want to get too distracted from the original goal of this PR (having a base schema for the current dcat-ap 2 profile). If we can come up with a better way of extending profiles, by all means let's explore it.

After looking at the actual changes, my gut feeling is that the HVD support can be directly incorporated in the current profile for parsing/serializing without needing a separate one, and the scheming bits can be grouped in a preset (or two, one for dataset and one for resource fields). 
Have you implemented GeoDCAT support as well or it was just an example? Would be good to see different approaches for this

EricSoroos:
We're reviewing Geo as well. (and a couple others, and localization)

Realistically, HVD is small except for some of the out of our scope spec issues (e.g., permanent identifiers, legislation in the MS to indicate that it's a HVD). It's also already included, so splitting it out involves some compatibility issues. 

Geo is in a similar space -- it's been included forever so if we do add a separate profile for it, it would be for Shackq compliant (existing extension) backwards incompatible changes.

amercader:
I think this is now ready to go, any further work should be done in separate PRs as this has grown quite a lot.

Highlights are:

* [Updated docs](https://github.com/ckan/ckanext-dcat/pull/281/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5) (I think it's time to set up proper RTD docs as the README is massive)
* Schema definitions ([all files](https://github.com/ckan/ckanext-dcat/pull/281/files#diff-77ebe48c2b442fff506a3aad288a513d3f116ded7c6f1eea67143f8e1abadb27) and only [recommended ones](https://github.com/ckan/ckanext-dcat/pull/281/files#diff-cb88d7e03df74be2c6e0087e17a9b97232adbb145523857940f34a0b9247b36f))
* [Custom validators](https://github.com/ckan/ckanext-dcat/pull/281/files#diff-7306a449cf228d8de8f92536423ec5406c40af284fce0255726dd89c84e5345e)
* Custom [indexing logic](https://github.com/ckan/ckanext-dcat/pull/281/files#diff-4c87ce47fea8818abd34d193928a7edccfb8c0db7e1a8d3cb6dd1e5836186962R165)
* The [compatibility profile](https://github.com/ckan/ckanext-dcat/pull/281/files#diff-4c87ce47fea8818abd34d193928a7edccfb8c0db7e1a8d3cb6dd1e5836186962R165) that tweaks the outputs of the existing profiles to support scheming definitions

ckanext/dcat/schemas/dcat_ap_2.1_full.yaml
```diff
@@ -0,0 +1,377 @@
+scheming_version: 2
+dataset_type: dataset
+about: A reimplementation of the default CKAN dataset schema
```
wardi, line 3:
```suggestion
about: Full DCAT AP 2.1 schema
```

ckanext/dcat/schemas/dcat_ap_2.1_recommended.yaml
```diff
@@ -0,0 +1,147 @@
+scheming_version: 2
+dataset_type: dataset
+about: A reimplementation of the default CKAN dataset schema
```
wardi, line 3:
```suggestion
about: Recommended fields for DCAT AP 2.1 schema
```

wardi:
This looks good.

I would be tempted to put more of the logic in the schemas but this extension needs to maintain backwards compatibility and ckanext-scheming-less operation so your approach makes sense.

ckanext/dcat/validators.py
```diff
@@ -0,0 +1,134 @@
...
+
+    if is_year(value) or is_year_month(value) or is_date(value):
+        return
+
+    try:
+        scheming_isodatetime({}, {})(key, data, errors, context)
+    except Invalid:
+        raise Invalid(
+            _(
+                "Date format incorrect. Supported formats are YYYY, YYYY-MM, YYYY-MM-DD and YYYY-MM-DDTHH:MM:SS"
```
EricSoroos, line 57: This is overly restrictive. The first datetime I attempted to parse with the dcat harvester was of the form: `YYY-MM-DDTHH:MM:SS.000Z`, which is permitted by ISO8601 (https://en.wikipedia.org/wiki/ISO_8601) and the `xsd:dateTime` (https://www.w3.org/TR/xmlschema11-2/#dateTime) spec. 

It appears that this goes back to the ckan helper  date_str_to_datetime which "Converts an ISO _like_ date to a datetime", using a 12 yr old regex based time zone ignoring date parser. This predates python having reasonable sensible timezone aware datetimes and the `datetime.datetime.fromisoformat` function. 

The core helper probably should be fixed, with its potential for knock on changes, but in the meantime since this is new code, perhaps we should just directly use `datetime.datetime.fromisoformat` here. (python3.7 min for that). We'd still need to support `YYYY` and `YYYY-MM` specially in the code, because those aren't covered

amercader: [`datetime.datetime.fromisoformat()`](https://docs.python.org/3/library/datetime.html#datetime.datetime.fromisoformat) is quite limited up until Python 3.11:

> Changed in version 3.11: Previously, this method only supported formats that could be emitted by [date.isoformat()](https://docs.python.org/3/library/datetime.html#datetime.date.isoformat) or [datetime.isoformat()](https://docs.python.org/3/library/datetime.html#datetime.datetime.isoformat).

So on python 3.10 and lower these dates are not parsed:

```
Python 3.10.10 (main, Mar 14 2023, 15:55:23) [GCC 9.4.0]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.13.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import datetime

In [2]: datetime.datetime.fromisoformat("2011-11-04T00:05:23Z")
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[2], line 1
----> 1 datetime.datetime.fromisoformat("2011-11-04T00:05:23Z")

ValueError: Invalid isoformat string: '2011-11-04T00:05:23Z'
```

What do you think of the approach I followed in https://github.com/ckan/ckanext-dcat/pull/281/commits/c7b8c027495f4dece84562090b44fd20f1374604? If something is not an `xsd:gYear`, an `xsd:gYearMonth` or an `xsd:date` we just let dateutil parse it (which will accept the timezone values you suggested). If that parses we serve it as an `xsd:dateTime`.
The tests check that 1) the input value is unchanged and 2) it's served with the correct `xsd` data type.

EricSoroos: Looks better than doing it ourselves. You'r enot testing for invallid dates, or ... conditionally valid dates like M/D/Y, but overall, I'm definitely happier with getting a builtin date parser that handles the in the wild formats that I've seen.

amercader: Added tests in [39b4d91](https://github.com/ckan/ckanext-dcat/pull/281/commits/39b4d9162c30772b4c10db17ee7126b734c95319)

amercader:
Merging this big chunk of changes, any followup like #288 we can do in smaller PRs

PR #282, opened 2024-05-23
Split profile classes into their own separate files

amercader:
The old profiles.py file was starting to become unwieldy. This messes up history but I think it's worth doing, considering there will be more profiles coming in the future. Also took the opportunity to format the new files with black


amercader:
Note: failures with master are caused by an issue with the `responses` package: https://github.com/getsentry/responses/issues/708, there should be a release soon: https://github.com/getsentry/responses/issues/714

PR #284, opened 2024-05-29
DCAT HVD 2.2.0

EricSoroos:
Draft version of HVD 2.2.0 compliance

1) Adds support for codelists as a concept, and uses them for the HVD Category
2) Uses appropriate rdf typing to pass the shaql validation.
3) Provides a read-only profile to provide a HVD only catalog at `{base url}/catalog.ttl?profiles=euro_dcat_ap_hvd_220`
4) Extracts resource_to_graph into overridable methods.

Todo:
* [ ] Shaql validation in automatic testing

### DIFF
--- a/ckanext/dcat/codelists/__init__.py
+++ b/ckanext/dcat/codelists/__init__.py
@@ -1,0 +1,8 @@
+
+from .extract import extract
+from pathlib import Path
+
+high_value_dataset_category = extract(Path(__file__).parent / 'high-value-dataset-category.rdf')
+
+
+__all__ = ['high_value_dataset_category']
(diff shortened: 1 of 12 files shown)


EricSoroos:
This is a draft for comment -- 

* I'm not clear that I'm hooking into the best place to provide the HVD only catalog. 
* The resource refactoring is pretty severe to just be able to drop resources out of the graph, there might be a better way. 
* I'm also not clear that a narrow subclass is a great idea to provide a minimized catalog, because if it's configured as a default profile, it would remove all non-hvd datasets from the catalog.

PR #285, opened 2024-06-04
Parse dcat:spatialResolutionInMeters as float

amercader:
According to the spec this should be typed as an xsd:Decimal, but we are parsing it as int:

https://www.w3.org/TR/vocab-dcat-3/#Property:dataset_spatial_resolution

WDYT @seitenbau-govdata ?

This allows to retrieve the correct values for decimal numbers (eg `1.5` instead of `1`). This changes slightly the values returned for integers (eg `1` to `1.0`) but as this function is only currently used in this property I think it's safe to introduce the change.

### DIFF
--- a/ckanext/dcat/profiles/base.py
+++ b/ckanext/dcat/profiles/base.py
@@ -230,13 +230,31 @@ class RDFProfile(object):
 
         Both subject and predicate must be rdflib URIRef or BNode objects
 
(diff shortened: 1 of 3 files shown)


seitenbau-govdata (approved):
@amercader Thanks for the improvement. The changes LGTM. 😃

amercader:
Great! thanks

Issue #249, opened 2023-06-01
two many locn:geometry


jze:
For DCAT-AP the extension generates too many `locn:geometry` values. 

DCAT-AP specifies the cardinality of `locn:geometry` for the class `dct:Location` as `[0..1]`. 

Currently the export looks like this

```xml
<dct:spatial>
  <dct:Location rdf:nodeID="N1c28db5fb3c64004809faa27ab9f2c18">
    <locn:geometry rdf:datatype="https://www.iana.org/assignments/media-types/application/vnd.geo+json">{"type":"Polygon","coordinates":[[[8.63647,54.30695],[8.63629,54.30572],[8.63691,54.30567],[8.63697,54.30688],[8.63647,54.30695]]]}</locn:geometry>
    <locn:geometry rdf:datatype="http://www.opengis.net/ont/geosparql#wktLiteral">POLYGON ((8.6365 54.3070, 8.6363 54.3057, 8.6369 54.3057, 8.6370 54.3069, 8.6365 54.3070))</locn:geometry>
  </dct:Location>
</dct:spatial>
```

I suggest to remove the GeoJSON entry. Then the export would be:

```xml
<dct:spatial>
  <dct:Location rdf:nodeID="N1c28db5fb3c64004809faa27ab9f2c18">
    <locn:geometry rdf:datatype="http://www.opengis.net/ont/geosparql#wktLiteral">POLYGON ((8.6365 54.3070, 8.6363 54.3057, 8.6369 54.3057, 8.6370 54.3069, 8.6365 54.3070))</locn:geometry>
  </dct:Location>
</dct:spatial>
```

amercader:
Fixed in https://github.com/ckan/ckanext-dcat/commit/83d1b6582870937fc5258d4977dcdc3c9332e4e7, part of https://github.com/ckan/ckanext-dcat/pull/288

PR #220, opened 2022-06-10
Add support for DCAT-AP 2.1

seitenbau-govdata:
We want to add support for the following DCAT-AP 2.1 properties (covers https://github.com/ckan/ckanext-dcat/issues/168):
* dcat:Dataset
  * dcat:bbox und dcat:centroid (in dct:Location of dct:spatial)
  * dcat:startDate, dcat:endDate, time:hasBeginning, time:hasEnd (in dct:PeriodOfTime of dct:temporal)
  * dcat:spatialResolutionInMeters
  * dcat:temporalResolution
  * dct:isReferencedBy
  * ~~prov:wasGeneratedBy~~
* dcat:Distribution
  * dcatap:availability
  * dcat:compressFormat
  * dcat:packageFormat
  * ~~odrl:hasPolicy~~

Specification: https://www.w3.org/TR/vocab-dcat-2/

We start by creating a new DCAT-AP 2 profile (plugin `euro_dcat_ap_2`) that inherits the profile `EuropeanDCATAPProfile` (plugin `euro_dcat_ap`).

PR #288, opened 2024-06-19
SHACL validation for DCAT-AP 2.1.1

amercader:
This includes #281 , which should be merged first.

As a next step after finishing the scheming support we will test against the official [SHACL](https://en.wikipedia.org/wiki/SHACL) shapes for [DCAT-AP 2.1.1](https://github.com/SEMICeu/DCAT-AP/tree/master/releases/2.1.1), to set up a reusable test mechanism and iron out existing issues with the current profiles.


amercader:
Right now, when testing against `dcat-ap_2.1.1_shacl_shapes.ttl` the only things left failing are the two values for the various geometry properties for dct:spatial (lcon:geometry, dcat:bbox, dcat:centroid). We are adding [both the GeoJSON and the WKT](https://github.com/ckan/ckanext-dcat/blob/83495ba99cba17398ba8feb1bc0da486f3798584/ckanext/dcat/profiles/base.py#L698-717) version of the geometry, and the validator doesn't like that. (cc @seitenbau-govdata )

amercader:
@seitenbau-govdata Back in #220 we introduced changes to expose geometries for LOCN.geometry, DCAT.bbox
or DCAT.centroid. The same geometry was serialized as GeoJSON and as WKT, adding a node for each format. This is not allowed according to the Shacl validation, so if we want to be compliant we need to output just one. I defaulted to WKT from now on as this is what GeoDCAT-AP requires (or GML yuk...), but sites that for some reason require GeoJSON (or both) can use the
`ckanext.dcat.output_spatial_format` to choose which format to use. Does that sound reasonable? Of course this will be noted on the changelog along other minor changes in the parsers.

seitenbau-govdata:
> @seitenbau-govdata Back in #220 we introduced changes to expose geometries for LOCN.geometry, DCAT.bbox or DCAT.centroid. The same geometry was serialized as GeoJSON and as WKT, adding a node for each format. This is not allowed according to the Shacl validation, so if we want to be compliant we need to output just one. I defaulted to WKT from now on as this is what GeoDCAT-AP requires (or GML yuk...), but sites that for some reason require GeoJSON (or both) can use the `ckanext.dcat.output_spatial_format` to choose which format to use. Does that sound reasonable? Of course this will be noted on the changelog along other minor changes in the parsers.

@amercader Thanks for the work on this! 👍 The serialization as GeoJSON and WKT was already there as we added the serialization for DCAT.bbox and DCAT.centroid and we decided to handle them in the same way. We were already aware of this, but decided not to change it due to backwards compatibility. This has already been reported with https://github.com/ckan/ckanext-dcat/issues/249.

But I think the time is ready now. 😃 And your suggestion with the parameter looks good. We would also prefer WKT for serialization. This should be fine for the most consumers, because the most libraries can handle several types of spatial data, e.g. GeoJSON and WKT. And the possibility to choose the serialization with `ckanext.dcat.output_spatial_format` should certainly cover all needs.
The current implementation for deserialization stores the spatial value as GeoJSON anyway. This will be kept, right?

PR #291, opened 2024-07-16
Add support for defining ranges/classes when generating a graph

amercader:
Starting off from some great work from @EricSoroos in #284, this PR modifies the base profile to support defining the range (`rdf:type`) of properties.

`_add_triple_from_dict()` & co now accept an optional `_class` parameter which is the RDF class that will get applied to the object, _if it is a URIRef_.  Classes can not be applied to Literals, otherwise you get invalid serializations.

I added a SHACL test using the [`dcat-ap_2.1.1_shacl_range.ttl`](https://github.com/SEMICeu/DCAT-AP/blob/master/releases/2.1.1/dcat-ap_2.1.1_shacl_range.ttl) shapes and we almost got to full validation but not quite. There were two classes that were not added to the profiles even if the range tests where failing:

* All the properties that have `skos:Concept`. Adding this class makes the range validation pass, but the shapes one fail, because we need to provide a `skos:prefLabel` value. This makes sense but the parser does not currently support it. It will probably have to be implemented as part of wider codelists support (which @EricSoroos also started on #284)
* All the properties that have `dcat:Dataset` range. Again, this makes the range tests pass but the shapes fail, because a `dct:title` and `dct:description` is expected of `dcat:Dataset`. I'm confused by this one, as it doesn't make sense to  require those for referenced datasets (things like `serves_dataset` or `is_version_of`) but maybe I'm missing something.

Both seem like not important enough to block the whole support for ranges, so until we find a way around it the range test ignores them


amercader:
@seitenbau-govdata @EricSoroos and other Semantic experts, would be good to get your take on this

PR #294, opened 2024-08-26
DCAT AP v3 profile

amercader:
New `euro_dcat_ap_3` profile with initial support for [DCAT AP v3](https://semiceu.github.io/DCAT-AP/releases/3.0.0). It includes changes in the `dcat:Dataset` and `dcat:Distribution` properties.

The v3 changes are not big in themselves but the profile classes have been refactored to a more "composition" approach to make easier to extend them going forward. For each DCAT AP version there are three kinds of methods: 

* `_base()` methods with common properties to *all* versions
* `_vX()` methods with properties introduced in this version and also applied to higher versions
* `_vX_only()` methods with properties only applied to the current version

Tested the serializations against the [official SHACL shapes](https://github.com/SEMICeu/DCAT-AP/tree/master/releases/3.0.0/shacl) (we are ignoring the same violations described in https://github.com/ckan/ckanext-dcat/pull/291for now)


amercader:
Anyone interested in trying the DCAT AP v3 profile, this is ready to give it a go

Markus92:
Great work @amercader !

Are there any plans for DatasetSeries support? It seems one of the most exciting (and difficult?) features of DCAT-3!

PR #301, opened 2024-09-09
Add dct identifier to publisher

hcvdwerf:
The system previously lacked support for storing and serializing publisher identifiers, such as ROR IDs, ORCIDs, or other unique identifiers for publishers. While other metadata fields like publisher name, email, and URL were being correctly parsed and serialized, the identifier field (e.g., dct:identifier) was not being included in the publisher’s metadata. This omission resulted in incomplete RDF metadata and affected datasets that needed to capture a unique reference to the publisher.


ckanext/dcat/converters.py
```diff
@@ -27,8 +27,17 @@ def dcat_to_ckan(dcat_dict):
     if isinstance(dcat_publisher, basestring):
         package_dict['extras'].append({'key': 'dcat_publisher_name', 'value': dcat_publisher})
     elif isinstance(dcat_publisher, dict) and dcat_publisher.get('name'):
-        package_dict['extras'].append({'key': 'dcat_publisher_name', 'value': dcat_publisher.get('name')})
-        package_dict['extras'].append({'key': 'dcat_publisher_email', 'value': dcat_publisher.get('mbox')})
+        if dcat_publisher.get('name'):
```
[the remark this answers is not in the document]
hcvdwerf: I removed the second one

ckanext/dcat/schemas/dcat_ap_recommended.yaml
```diff
@@ -66,6 +66,11 @@ dataset_fields:
 
     - field_name: type
       label: Type
+
+    - field_name: identifier
```
amercader, line 70: Can you add it to the dcat_ap_full.yaml file as well?

ckanext/dcat/profiles/euro_dcat_ap_base.py
```diff
@@ -123,7 +123,7 @@ def _parse_dataset_base(self, dataset_dict, dataset_ref):
 
         # Publisher
         publisher = self._publisher(dataset_ref, DCT.publisher)
-        for key in ("uri", "name", "email", "url", "type"):
+        for key in ("uri", "name", "email", "url", "type", "identifier"):
```
amercader, line 126: This will take care of the DCAT -> CKAN mapping, but you also need to add the other way around, CKAN -> DCAT. This is done a bit further down in the file [here](https://github.com/ckan/ckanext-dcat/blob/cb25389f11effefc28165b54de9ec68f61a86898/ckanext/dcat/profiles/euro_dcat_ap_base.py#L370) and [here](https://github.com/ckan/ckanext-dcat/blob/cb25389f11effefc28165b54de9ec68f61a86898/ckanext/dcat/profiles/euro_dcat_ap_base.py#L404).  

Additionally, this will cover the legacy way of defining publisher fields based on namespaced `publisher_*` extra fields, but going forward we want to support scheming objects (`"publisher": {"id": xx, "name": yy, "url": zz}`). This is handled in the `euro_dcat_ap_scheming` profile [here](https://github.com/ckan/ckanext-dcat/blob/cb25389f11effefc28165b54de9ec68f61a86898/ckanext/dcat/profiles/euro_dcat_ap_scheming.py#L169) (for CKAN -> DCAT, the opposite should already be handled by the base profile). Do you mind adding support for the new field there as well? And if you extend [this test](https://github.com/ckan/ckanext-dcat/blob/cb25389f11effefc28165b54de9ec68f61a86898/ckanext/dcat/tests/profiles/dcat_ap_2/test_scheming_support.py#L303) to cover the new field, even better.

@hcvdwerf Let me know if all this makes sense

hcvdwerf: Tnx for the comments. Verry helpful @amercader . Jut commit the changes

Markus92 (changes requested):
Small fix for failing e2e test.

amercader:
Thanks @hcvdwerf !

Issue #307, opened 2024-09-26
Support for multiple contact points/publishers


hcvdwerf:
DCAT-AP 3.0 makes the inclusion of multiple contacts mandatory for datasets. Specifically, the dcat:contactPoint property must be able to reference multiple contact points. This requirement aligns with the latest DCAT-AP specification. 

Code I use to test 

```
@prefix dcterms: <http://purl.org/dc/terms/> .
@prefix dcat: <http://www.w3.org/ns/dcat#> .
@prefix foaf: <http://xmlns.com/foaf/0.1/> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
@prefix ldp: <http://www.w3.org/ns/ldp#> .

<https://health-ri.sandbox.semlab-leiden.nl/dataset/d7129d28-b72a-437f-8db0-4f0258dd3c25>
  a dcat:Resource, dcat:Dataset;
  <http://www.w3.org/2000/01/rdf-schema#label> "Example";
  dcterms:title "Example";
  <https://w3id.org/fdp/fdp-o#metadataIssued> "2023-09-05T12:00:36.276171042Z"^^xsd:dateTime;
  <https://w3id.org/fdp/fdp-o#metadataModified> "2024-05-02T13:01:35.716385359Z"^^xsd:dateTime;
  dcterms:license <http://rdflicense.appspot.com/rdflicense/cc-by-nc-nd3.0>;
  dcterms:description "This is an example description.";
  dcat:contactPoint <https://orcid.org/0000-0002-9095-9201>, <https://orcid.org/0000-0003-2558-7496> .
```

```
@pytest.mark.ckan_config("ckan.plugins", "scheming_datasets")
@pytest.mark.usefixtures("with_plugins")
def test_profile_contact_point_multiple_uris():
    fdp_record_to_package = FairDataPointRecordToPackageConverter(profile="fairdatapoint_dcat_ap")
    data = Graph().parse(Path(TEST_DATA_DIRECTORY, "contact_point_multiple_urls.ttl")).serialize()
    actual = fdp_record_to_package.record_to_package(
        guid="https://health-ri.sandbox.semlab-leiden.nl/catalog/e3faf7ad-050c-475f-8ce4-da7e2faa5cd0;"
             "dataset=https://health-ri.sandbox.semlab-leiden.nl/dataset/d7129d28-b72a-437f-8db0-4f0258dd3c25",
        record=data)
    expected = {
        'extras': [],
[...]

PR #310, opened 2024-09-27
fix(extras): handle multiple contacts and publishers with 1-based ind…

hcvdwerf:
…exing

- Adjust contact and publisher handling to support multiple entries with 1-based indexing
- Append contact details with keys like `contact_email_1`, `contact_uri_1`, etc.
- Append publisher details with keys like `publisher_uri_1`, `publisher_name_1`, etc.
- Ensure consistent indexing when storing contact and publisher information
- Closes issue #307


amercader:
See https://github.com/ckan/ckanext-dcat/issues/307#issuecomment-2385313502

PR #312, opened 2024-10-03
Solves URI from contact point

hcvdwerf:
#### Description:
This pull request introduces functionality to resolve URIs for contact details, specifically targeting the retrieval of more detailed information (e.g., ORCID). Previously, the `contact["uri"]` field defaulted to either a string or an empty value. With this update, we attempt to resolve the URI and retrieve associated metadata (e.g., `VCARD.hasUID`) when available.

#### Changes:
- Updated the `_contact_details` method to resolve URIs where applicable.
- For agents that are not `URIRef`, the system now attempts to fetch the `VCARD.hasUID` property to enrich the URI field.

#### Benefits:
- **Improved Metadata Access**: Resolving URIs, such as ORCIDs, allows us to retrieve richer and more accurate contact data (e.g., full names, affiliations, email addresses).


amercader:
Wouldn't it make more sense to store `VCARD.hasUID` in a dedicated key like `contact["identifier"]`. I get that in the case of ORCID those happen to be URIs as well but that might not be always be the case right? e.g. if using a UUID

ckanext/dcat/profiles/base.py
```diff
@@ -465,7 +465,8 @@ def _contact_details(self, subject, predicate):
 
         for agent in self.g.objects(subject, predicate):
 
-            contact["uri"] = str(agent) if isinstance(agent, term.URIRef) else ""
+            contact["uri"] = (str(agent) if isinstance(agent, URIRef)
+                        else self._get_vcard_property_value(agent, VCARD.hasUID))
```
amercader, line 469: Let's keep this as it was and not assign hasUID to `uri` as discussed. 
Otherwise this is good to go

hcvdwerf: I reverted this part back and added also the new mapping to the documentation

PR #322, opened 2024-11-19
fix: Move has version from distribution to dataset

hcvdwerf:
This pull request corrects the placement of the dcat:hasVersion property in our DCAT implementation. Previously, the property was mistakenly defined under Distribution instead of Dataset. This update ensures compliance with the DCAT specification by associating dcat:hasVersion with Dataset.

I must admit, this was my mistake @amercader


hcvdwerf:
@amercader any idea why tests are failing?

amercader:
@hcvdwerf tests failures are unrelated, they come from the docker images: https://github.com/ckan/ckan-docker-base/issues/86

PR #324, opened 2024-12-04
Add VCARD.hasURL property parsing/serialization

Markus92:
As part of our HealthDCAT-AP work we identified a need to add the URL property to `dcat:contactPoint`. It is defined in the [VCARD RDF ontology](https://www.w3.org/TR/vcard-rdf/) and meant to link to a contact page for example.

I cherrypicked another small fix in this branch, which is that the `foaf:mbox` property does not strip the *mailto:* prefix, which is consistently stripped in the rest of the CKAN DCAT extension. It made a few of our test cases fail, I'll include those with the HealthDCAT-AP PR.

### DIFF
--- a/ckanext/dcat/harvesters/base.py
+++ b/ckanext/dcat/harvesters/base.py
@@ -159,17 +159,12 @@ class DCATHarvester(HarvesterBase):
         '''
         Returns a database result of datasets matching the given guid.
         '''
-        if toolkit.check_ckan_version(max_version="2.11"):
-            datasets = model.Session.query(model.Package.id) \
-                                    .join(model.PackageExtra) \
-                                    .filter(model.PackageExtra.key == 'guid') \
-                                    .filter(model.PackageExtra.value == guid) \
-                                    .filter(model.Package.state == 'active') \
-                                    .all()
-        else:
-            datasets = model.Session.query(model.Package.id) \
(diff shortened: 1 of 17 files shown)


ckanext/dcat/harvesters/base.py
```diff
@@ -159,17 +159,12 @@ def _read_datasets_from_db(self, guid):
         '''
         Returns a database result of datasets matching the given guid.
         '''
-        if toolkit.check_ckan_version(max_version="2.11"):
```
amercader, line 162: This was probably changed by mistake

amercader (commented):
Apart from the unrelated changed mentioned in the comment this looks good to go, thanks @Markus92

ckanext/dcat/harvesters/base.py
hcvdwerf: Unfortunately, when I revert this setting, the pipeline fails. Please refer to the following link for details: https://github.com/ckan/ckanext-dcat/actions/runs/12284575222. The tests seem to fail unexpectedly.

hcvdwerf: Maybe something upstream changed?

amercader: My bad, the version check was wrong: https://github.com/ckan/ckanext-dcat/commit/6bcf6931158ce995487a0828accf926ecfb375b1

Issue #315, opened 2024-10-15
HealthDCAT-AP Profile


Markus92:
The European Parliament recently passed the [European Health Data Space (EHDS)](https://health.ec.europa.eu/ehealth-digital-health-and-care/european-health-data-space_en) act. As part of the EHDS, all member states will set up infrastructure for the re-use of health data. Part of this is the [HealthDCAT-AP profile](https://healthdcat-ap.github.io/), currently in draft. Once finalized, member states will have to implement support for this profile into their own catalogues.

As the Dutch national Health Research Infrastructure, we at Health-RI would also implement this on top of our CKAN catalog. We were wondering if CKAN would be open to us upstreaming this specific profile?  From a maintenance and sustainability perspective we highly prefer upstreaming whenever possible, and it would make it a lot easier for our colleagues in the rest of the EU! But we also understand this might be a bit of a niche use case.

amercader:
@Markus92 this sounds really interesting. I'm keen on other organizations building custom profiles on top of the base ones provided by the extension to identify what works and what doesn't. So I'll take this as an opportunity to provide some wider guidance on how to contribute profiles to the extension.

I'm definitely open to the Health profile being upstreamed to this extension, but we should make sure that anything that is merged is sustainable long term. This means that if in one or two years neither you as the original developers or myself as a maintainer are around, the maintainers at the time can 1) know what is that about and 2) still be confident that everything works as expected. The way to achieve this is with tests (lots of them!) and documentation.

This is I'd like to see in a contribution that adds support for a new profile:

* A new [profile class](https://github.com/ckan/ckanext-dcat/tree/master/ckanext/dcat/profiles) with parse and serialize methods (extending the DCAT v3) one
* A new dataset [Example](https://github.com/ckan/ckanext-dcat/tree/master/ckanext/dcat/schemas) that contains all new properties supported in the new profile (it can contain just the base DCAT 3 recommended ones)
* [Example](https://github.com/ckan/ckanext-dcat/tree/master/examples) CKAN dataset and DCAT serialization of the new profile
* Tests:
    * [SHACL validation](https://github.com/ckan/ckanext-dcat/tree/1e945b6e79f0e0bae1ff76989ef9789abb5e32a8/ckanext/dcat/tests/shacl) if SHACL shapes are provided
[...]

PR #328, opened 2024-12-17
Add support for the Croissant metadata specification

Reikyo:
Added namespace, schema and profile for the Croissant metadata specification ([https://docs.mlcommons.org/croissant/docs/croissant-spec.html](https://docs.mlcommons.org/croissant/docs/croissant-spec.html))


Reikyo:
Hi @amercader - I'm working with MLCommons on a project being run by @benjelloun at Google, aiming to integrate capability for the Croissant metadata specification as widely as possible.

Over the last few weeks we've been exploring CKAN and its extension ecosystem, and found that the existing ckanext-dcat extension already provides the needed functionality. I've created a new branch with a single commit showing all required adjustments. To existing files there are only changes to the list of namespaces, and to the list and import of profiles. The new schema and profile are in new separate files isolated from the existing logic, and have been made according to the documentation ([schemas](https://docs.ckan.org/projects/ckanext-dcat/en/latest/getting-started/#schemas), [profiles](https://docs.ckan.org/projects/ckanext-dcat/en/latest/profiles/)) and using the existing schemaorg.py profile as a basis.

You can find some further conversation on this topic [here](https://github.com/ckan/ckan/discussions/8414). I'd be grateful if you can review the PR and consider it for integration into master. Please let me and @benjelloun know if you have any questions.

Many thanks for your help.

amercader:
@Reikyo this is great and extremely timely as I had started working myself on a croissant profile these last few days. I think this is an extremely valuable feature for sites that we can get ready relatively soon. I'll review asap and get back to you.

Regarding where this functionality should live I think it wouldn't be a big task to separate the schema.org/croissant profiles to their own extension but let's focus on getting this ready for now.

Thanks, this is really exciting

Reikyo:
@amercader Thanks for the quick reply, much appreciated.

You can find some further info about the work done on this [here](https://drive.google.com/drive/folders/1HZS2x748myJje75ekomhfzi-RSSpHwUR). In particular, see the [spreadsheet](https://docs.google.com/spreadsheets/d/1kxjtfT6eFO-A3ZG3B2D73MTp4yh7lRYEPUdBrAZvdIs) for indication of the properties and values being considered, and how they are present in the Croissant RDF graph when no scheming schema is applied to modify the UI (column O) and when the new Croissant scheming schema is applied (column S). In the latter case there are, of course, many more output properties, which go up to (but not including) the [RecordSet](https://docs.mlcommons.org/croissant/docs/croissant-spec.html#recordsets) section of the Croissant specification. You'll find a number of notes throughout the new schema (schemas/croissant.yaml) and profile (profiles/croissant.py) to clarify the details and reasoning. Also note that for the schema, I allowed users to define their own "@id"s where applicable, and you can see these used in the profile where "id_given" is present.

In the above folder I've also included images and files that show what I'm thinking of as a three step process. There are files showing all three steps when no scheming schema is applied, and when the Croissant scheming schema is applied.

- Step-1: Data is input via edit forms (dataset and resource)
- Step-2: Data is represented in a Python dictionary (dataset_dict)
- Step-3: Data is output in RDF representation, which can be serialised in various ways (no/default profile, schemaorg profile, Croissant profile)

Hopefully this makes it clear exactly what the changes do.

amercader:
@Reikyo I had a first pass it looks like all this is going in the right direction.

Some points in no particular order:

* The new profile will translate the internal CKAN dict representation to a Croissant one but we still need a mechanism to expose that to users. From a user point of view, they should just enable a `croissant` plugin in their [`ckan.plugins`](https://docs.ckan.org/en/latest/maintaining/configuration.html#ckan-plugins) config option and forget about it. The plugin will take care of embedding the JSON-LD code in the dataset page. This is how the schema.org markup for Google Dataset Search works now (with the [`structured_data` plugin](https://docs.ckan.org/projects/ckanext-dcat/en/latest/google-dataset-search/)). So we need a plugin similar to [this one](https://github.com/ckan/ckanext-dcat/blob/875e69cc6f181a034c50f06a007afd8b1b400ee0/ckanext/dcat/plugins/__init__.py#L245) (registered in pyproject.toml) that adds a new template helper that embeds the code in the [`package/read_base.html`](https://github.com/ckan/ckanext-dcat/blob/875e69cc6f181a034c50f06a007afd8b1b400ee0/ckanext/dcat/templates/package/read_base.html#L30) template.
(*Aside:* the current code could be improved because `structured_data` relies on the `dcat` plugin being loaded, but I'll take care of that)

* Is the croissant JSON-LD representation meant to only be embedded in the HTML source of the dataset page? or is there value in having it available at a dedicated endpoint (e.g. `/dataset/<id>/croissant.jsonld`). I see Hugging Face does both so I imagine there is value. In any case it should be easy to expose an endpoint, using [`IBlueprint`](https://github.com/ckan/ckanext-dcat/blob/875e69cc6f181a034c50f06a007afd8b1b400ee0/ckanext/dcat/plugins/__init__.py#L73)

* It's great that you have mapped the properties between the standard CKAN model properties (i.e. no custom schema, column O) and added a new schema that adds the ones not present. In that case I think it's important that the properties follow the existing names in the DCAT profiles, so sites can get metadata exposed in both formats using a single schema. Going through the list I believe this is already the case but flagging just in case.

* At first I was wondering if `id_given` might be comparable to an URI and so if we could keep `uri` as the field name for consistency with the DCAT profiles but looking at the spec and your comments in the schema it seems like there are clear conventions and recommendations for defining ids in the context of Croissant so it make sense to keep these separate.

* I'll take care of putting these properly in the docs, but [here's a summary](https://github.com/ckan/ckanext-dcat/issues/315#issuecomment-2443814392) of what we would need a new profile to have to include it in this extension (basically to ensure that is properly tested and documented). We have the profile class and the schema, so what's missing is:
  * Examples
  * Tests, in this case we only need serialization tests (e.g. given a certain CKAN dataset in the examples folder we get the expected croissant properties, see the [schema.org ones](https://github.com/ckan/ckanext-dcat/blob/master/ckanext/dcat/tests/profiles/schemaorg/test_schemaorg_profile_serialize.py)) and we could also validate the output with the [python croissant package](https://github.com/ckan/ckanext-dcat/blob/master/ckanext/dcat/tests/profiles/schemaorg/test_schemaorg_profile_serialize.py) to make sure there are no warnings.
  * Docs: brief intro of what the Croissant spec is, what the plugin does and how to enable it.

Let me know if all this makes sense. I can help with any of the points above, in fact I just pushed a new `croissant` branch on this repo. If you change the target of this PR to that branch I can push stuff there as well to help move forward quicker if it helps.

I'll be mostly off the next couple of weeks though, but can pick it up again in January.

Reikyo:
Moving the conversation from here to [https://github.com/ckan/ckanext-dcat/pull/330](https://github.com/ckan/ckanext-dcat/pull/330).

PR #339, opened 2025-02-14
Support for Croissant ML

amercader:
This PR compiles all work done in #328 in #330 by @Reikyo and myself to add support for the [Croissant ML](https://mlcommons.org/working-groups/data/croissant/) specification.

A new profile serializes dataset dicts following the [Croissant spec](https://docs.mlcommons.org/croissant/docs/croissant-spec.html). This serialization is embedded in the dataset page source and also available in a dedicated endpoint (`/dataset/<dataset-name>/croissant.jsonld`)

The new documentation is available here: https://docs.ckan.org/projects/ckanext-dcat/en/croissant/croissant/

TODO:
- [x] Decide whether to use framing or not for the JSON-LD output
- [x] Decide whether to include `"@language": "en"` in the context
- [x] Example croissant output
- [x] Tests for serialization and for custom endpoint

In parallel I've been working on adding support for [RecordSet](https://docs.mlcommons.org/croissant/docs/croissant-spec.html#recordsets) objects but that will go into a separate PR


amercader:
@Reikyo this is good to go AFAIC

Writing tests made surface some minor issues, and I finally decided to drop the `@language` property in the context, not only to make the output clearer but we also can not assume what the language of the metadata in different CKAN sites will be so it's better to let users create their custom profiles with whatever makes sense.

Let me know if you think anything else is needed, otherwise I'll merge this.

I'll follow soon with a PR to add support for `RecordSet` objects.

Reikyo:
@amercader Thanks, I've had a look at recent changes and can suggest some tweaks. Firstly, I saw that you adjusted from "title" to "name" in a couple of places in the schema. But as per the [note](https://github.com/ckan/ckanext-dcat/blob/croissant/ckanext/dcat/schemas/croissant.yaml#L199) in the first instance, the original version was in order to pick up `organization["title"]` in the [fallback case for creator](https://github.com/ckan/ckanext-dcat/blob/croissant/ckanext/dcat/profiles/croissant.py#L205). The `organization` dict has "title" and "name" fields, the latter of which is actually its URL slug, similar to how "name" is used for the main part of the dataset dict itself. As the new code currently stands, the fallback case now doesn't work as intended [here](https://github.com/ckan/ckanext-dcat/blob/croissant/ckanext/dcat/profiles/croissant.py#L239). This could all be propped up with some further conditional logic in the profile, or by reverting to the original use of "title" instead of "name" in the schema for both creator and publisher. What do you think?

Secondly, I'll have a thorough look through the documentation examples later this afternoon, but one thing I already noticed was that there's a hash for one of the subresources. Following our discussion about this recently, we now know that there's a hash for the top-level resource only, so the subresource one has to be removed.

Reikyo:
@amercader Sorry, just saw you also added `agent_dict["name"] = agent_dict["title"]` to cater for the fallback. So let's leave it as that, but I'll remove the comments in the schema which are now obsolete.

amercader:
@Reikyo re orgs, I was going to link to that line, but I added a test in [a79d0aa](https://github.com/ckan/ckanext-dcat/pull/339/commits/a79d0aabfaec1f54324b7414a14a631ec1cc939d) anyway.

re hashes in sub-resources, should be fixed in [1273cac](https://github.com/ckan/ckanext-dcat/pull/339/commits/1273cac06f05708036e099d53d3fad8d2018e5cc), and I also updated the example.

Reikyo:
@amercader I've made two further commits to my version, [cc34b6c](https://github.com/ckan/ckanext-dcat/commit/cc34b6c5bc2aa7dfc02f8445f80370f66cdabf59) and [c365bd5](https://github.com/ckan/ckanext-dcat/commit/c365bd59cfdf0829e251330789ef24d5dfca89d8). As I was working on this simultaneously to you, I had an intermediate merge to sync with your changes, you can see that sandwiched [here](https://github.com/ckan/ckanext-dcat/compare/croissant...Reikyo:ckanext-dcat:croissant). Let me know if you want me to just submit a PR anyway into this central version. Seeing as you already had this PR open to merge into master, I thought you may want to apply the diffs yourself into the set of commits for this PR instead.

amercader:
@Reikyo I think this is good to merge. Any further changes we can do in separate PRs. Thanks a lot for your work on this!

Reikyo:
@amercader That's great, I'll pass on the news to the Croissant Working Group, they'll be very happy to hear that we now have another route for supporting the standard in the wider data/ML community. Thanks again for your help and support with this, your expertise has been crucial to ensure that it all meshes well with the existing infrastructure and becomes a native part of the extension. Very much appreciated!

amercader:
@Reikyo It's a great addition to the CKAN ecosystem, I'm planning on writing a small blog post on ckan.org around this so I might ping you for feedback. I'm also planning on submitting a PR to mlcommons/croissant to add CKAN to the "Integrations" section in the README. If you know of other places where this could be advertised let me know.

I want to try to merge #341 before doing a new ckanext-dcat release with this feature so if you have time to have a quick look that would be great!

Reikyo:
@amercader I don't have a PostgreSQL database currently set up to run DataStore and check this in practice, but I've looked at the changes to `croissant.py` and have inferred that the output should look something like the following:

```
"cr:recordSet": {
	"@id": "my-recordset-ref",
	"@type": "cr:RecordSet",
	"sc:name": "my-recordset-ref",
	"sc:field": [
		{
			"@id": "my-field-ref",
			"@type": "cr:Field",
			"cr:dataType": "my-datatype",
			"sc:source": {
				"sc:extract": {
					"sc:column": "my-field-id"
				},
				"sc:fileObject": {
					"@id": "my-resource-ref"
				}
			}
		}
	],
	"sc:key": [
		{
			"@id": "my-field-ref"
		}
	]
}
```

This is just what can be seen in `full_croissant_dataset.jsonld` and `croissant.md` in terms of structure, so if they're actually from a live run of the code then that's great. Conclusion is that I think this is good providing it's reading from DataStore as intended, but I'll defer to your expertise there.

A couple of tweaks that might be considered:

- Remove the "sc:name" field - this isn't in the Croissant spec, and it's just the same as the "@id" anyway from how the code works. Actually, I do notice one example of this in the Croissant spec now that I've checked ... so if it doesn't cause validation issues then maybe no problem.
- Even though the following vocabulary prefixes aren't apparent in the output, the Croissant spec context does indicate that they are Croissant properties rather than schema.org properties, in which case `croissant.py` should be changed as:
  - SCHEMA.field -> CR.field
  - SCHEMA.source -> CR.source
  - SCHEMA.extract -> CR.extract
  - SCHEMA.column -> CR.column
  - SCHEMA.fileObject -> CR.fileObject
  - SCHEMA.key -> CR.key

Thanks for going the extra mile with RecordSet!

PR #342, opened 2025-02-25
Include dependencies in pyproject, update install docs

### LINKED ISSUE #340: Manifest file should include requirements.txt
`MANIFEST.in` should include the `requirements.txt` file so that dependencies can be identified and installed without needing an editable install (`pip install -e`).

Alternatively, dependencies should be added to `pyproject.toml`

amercader:
Fixes #340 

Keep the requirements.txt and dev-requirements.txt around and acting like the source of truth, so all existing installation processes keep working as intended but include them dynamically in pyproject.toml to integrate with modern workflows, and to be able to just run:

    pip install ckanext-dcat

The requirements were loosen a bit to adapt to the suggestions by @smotornyuk here:

https://github.com/ckan/ckan/issues/8382#issuecomment-2353188970

### DIFF
--- a/.github/workflows/publish-pypi.yml
+++ b/.github/workflows/publish-pypi.yml
@@ -1,0 +1,54 @@
+name: Publish to PyPI
+
+# Publish to PyPI when a tag is pushed
+on:
+  push:
+    tags:
+      - 'v**'
+
+jobs:
+  build:
+    if: github.repository == 'ckan/ckanext-dcat'
+    name: Build distribution
+    runs-on: ubuntu-latest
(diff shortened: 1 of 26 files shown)


seitenbau-govdata:
Thanks @amercader ! That's pretty cool. We are working on updating to the latest version right now and encountered the same "issues" and solved them as you have now. So we are excitedly awaiting the next release and will update directly to it. 😃

amercader:
Then let's wait no more! I think all changes here are improvements and controversial so I'll do a new release now

PR #344, opened 2025-03-04
Fix TIME namespace definition to use DCAT recommendation

mjanez:
### Context
This PR addresses an issue with namespace definitions while working on implementing profiles according to the [contributing guidelines](https://github.com/ckan/ckanext-dcat/blob/f964c66030e1b81a181ff58d565aa529cadb3e4d/docs/contributing.md#including-new-profiles). We've been developing custom profiles for the Spanish application profiles - both the [current NTI-RISP (based on DCAT)](https://datos.gob.es/sites/default/files/doc/file/guia_nti_pdf_reutilizacion_recursos_informacion_2aed.pdf) and the future Spanish profile based on DCAT-AP.

Although this can be handled within the harvester, following DCAT's recommendation to use the more common namespace would be beneficial, especially since many publishers rely on ckanext-dcat to serialize their RDF for the national catalog.

### Problem
During development, we identified an issue with the TIME namespace definition in base.py. The current implementation defines TIME without the trailing hash (#):

```python
TIME = Namespace("http://www.w3.org/2006/time")
```

However, the correct URI according to W3C specifications and DCAT-AP should include the trailing hash:

```python
TIME = Namespace("http://www.w3.org/2006/time#")
```

### Impact
This incorrect namespace definition causes properties like `time:years`, `time:days`, etc. to be serialized with incorrect URIs:
- Current (incorrect): `http://www.w3.org/2006/timeyears`
- Expected (correct): `http://www.w3.org/2006/time#years`

This breaks interoperability with standards-compliant harvesters, especially for `dct:accrualPeriodicity` data structures for federation with portals like datos.gob.es.

### Example of the issue

We found datasets that weren't federating correctly with datos.gob.es. Examining the RDF revealed that years/days properties weren't being properly recognized:

**Incorrect (won't federate):**
```xml
<?xml version="1.0" encoding="utf-8"?>
<rdf:RDF
  xmlns:adms="http://www.w3.org/ns/adms#"
  xmlns:hydra="http://www.w3.org/ns/hydra/core#"
  xmlns:foaf="http://xmlns.com/foaf/0.1/"
  xmlns:dc="http://purl.org/dc/elements/1.1/"
  xmlns:dcat="http://www.w3.org/ns/dcat#"
  xmlns:locn="http://www.w3.org/ns/locn#"
  xmlns:vcard="http://www.w3.org/2006/vcard/ns#"
  xmlns:schema="http://schema.org/"
  xmlns:rdfs="http://www.w3.org/2000/01/rdf-schema#"
  xmlns:dct="http://purl.org/dc/terms/"
  xmlns:odrs="http://schema.theodi.org/odrs#"
  xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
  xmlns:time="http://www.w3.org/2006/time"
  xmlns:cnt="http://www.w3.org/2011/content#"
>
...

<dcat:Dataset rdf:about="https://example.org/catalogo/dataset/54336a93-2478-44fc-bb78-696c77cff5c2">
  <dct:accrualPeriodicity>
    <dct:Frequency rdf:nodeID="Ne107337f7f944ee0b67a3813d529b04f">
      <rdf:value>
        <time:DurationDescription rdf:nodeID="N5a50f9faffc542649ba2ba516a14a837">
          <rdfs:label>2 años</rdfs:label>
          <time:years rdf:datatype="http://www.w3.org/2001/XMLSchema#integer">2</time:years>
        </time:DurationDescription>
      </rdf:value>
    </dct:Frequency>
  </dct:accrualPeriodicity>

...
```

Here, `time:days` points to `http://www.w3.org/2006/timeyears` instead of `http://www.w3.org/2006/time#years`.

**Correctly federated example:**
```xml
<?xml version="1.0" encoding="utf-8"?>
<rdf:RDF
  xmlns:adms="http://www.w3.org/ns/adms#"
  xmlns:hydra="http://www.w3.org/ns/hydra/core#"
  xmlns:foaf="http://xmlns.com/foaf/0.1/"
  xmlns:dc="http://purl.org/dc/elements/1.1/"
  xmlns:dcat="http://www.w3.org/ns/dcat#"
  xmlns:locn="http://www.w3.org/ns/locn#"
  xmlns:vcard="http://www.w3.org/2006/vcard/ns#"
  xmlns:schema="http://schema.org/"
  xmlns:rdfs="http://www.w3.org/2000/01/rdf-schema#"
  xmlns:dct="http://purl.org/dc/terms/"
  xmlns:odrs="http://schema.theodi.org/odrs#"
  xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
  xmlns:time="http://www.w3.org/2006/time#"
  xmlns:cnt="http://www.w3.org/2011/content#"
>
...

<dcat:Dataset rdf:about="https://example.org/catalogo/dataset/54336a93-2478-44fc-bb78-696c77cff5c2">
  <dct:accrualPeriodicity>
    <dct:Frequency rdf:nodeID="Ne107337f7f944ee0b67a3813d529b04f">
      <rdf:value>
        <time:DurationDescription rdf:nodeID="N5a50f9faffc542649ba2ba516a14a837">
          <rdfs:label>2 años</rdfs:label>
          <time:years rdf:datatype="http://www.w3.org/2001/XMLSchema#integer">2</time:years>
        </time:DurationDescription>
      </rdf:value>
    </dct:Frequency>
  </dct:accrualPeriodicity>

...
```
### Solution
This PR updates the namespace definition in `base.py` to use the correct URI with the trailing hash, as specified in [DCAT](https://www.w3.org/TR/vocab-dcat-3/#normative-namespaces):

````diff
@@ -23,7 +23,7 @@
 VCARD = Namespace("http://www.w3.org/2006/vcard/ns#")
 FOAF = Namespace("http://xmlns.com/foaf/0.1/")
 SCHEMA = Namespace("http://schema.org/")
-TIME = Namespace("http://www.w3.org/2006/time")
+TIME = Namespace("http://www.w3.org/2006/time#")
 LOCN = Namespace("http://www.w3.org/ns/locn#")
 GSP = Namespace("http://www.opengis.net/ont/geosparql#")
 OWL = Namespace("http://www.w3.org/2002/07/owl#")
````

Also [update namespace in tests](https://github.com/ckan/ckanext-dcat/pull/344/commits/3f10de66a5218b4a9654e4193226ceb995345648)

### Testing & Verification
We've verified this fix by manually applying it to our local installation and confirming that datasets with temporal properties using the TIME ontology (e.g., `dct:accrualPeriodicity`) are now correctly harvested by external systems.

[Tests also pass](https://github.com/mjanez/ckanext-dcat/actions/runs/13674396803).

### References
- [W3C Time Ontology Specification](https://www.w3.org/TR/owl-time/)
- [DCAT Vocabulary](https://www.w3.org/TR/vocab-dcat-3/#normative-namespaces)
- [DCAT-AP Specification](https://semiceu.github.io/DCAT-AP/releases/3.0.0/#used-prefixes)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased](https://github.com/ckan/ckanext-dcat/compare/v2.3.0...HEAD)
 
+* Fix TIME namespace definition to include trailing hash (`#`), ensuring compliance with W3C Time Ontology and DCAT-AP. This may require updates for custom profiles relying on the old variant. ([#344](https://github.com/ckan/ckanext-dcat/pull/344))
 * Add missing URIRefOrLiteral import in profiles module. ([#343](https://github.com/ckan/ckanext-dcat/pull/343))
 
 ## [v2.3.0](https://github.com/ckan/ckanext-dcat/compare/v2.2.0...v2.3.0) - 2025-02-25

### CODE REVIEW

**Comment by mjanez** [comment]
Thanks @amercader

Good point, you're right that the example I gave with `time:years` might be confusing since it’s not from the vanilla `ckanext-dcat`.

The context here is an external system that consumes RDF serialization provided by a custom DCAT ES profile based on `ckanext-dcat`, and it inherited its own TIME namespace.

The specific case serializes the RDF with time periods using `time:DurationDescription` instead of the [URIRefOrLiteral serialization](https://github.com/ckan/ckanext-dcat/blob/3afc1e2edd59efaced33980ccabd0a7d86e96d77/ckanext/dcat/profiles/euro_dcat_ap_base.py#L348), which wasn’t compatible with the “current” application profile (pre-DCAT-AP), producing something like this:

```ttl
dct:accrualPeriodicity 
[ 
    a dct:Frequency; 
    rdf:value 
    [ 
        a time:DurationDescription; 
        rdfs:label "{time-interval}"; 
        time:{period} {n}. 
    ]; 
    rdfs:label "Every {time-interval}". 
]; 
```

The issue is that the harvesting extension defines its own TIME namespace, and it did have the `#`:

https://github.com/ctt-gob-es/datos.gob.es/blob/30c4a0d97356e0caf948aff2bb74790f4885c67f/ckan/ckanext-dge-harvest/ckanext/dge_harvest/profiles.py#L54

So, when a publisher generates RDF that doesn’t match the same TIME namespace declaration, it fails and throws an error. Because of the parsing method's quirks, it never parsed the update frequency.

Probably the only thing to check would be if all namespaces are used canonically, which is probably the case since this is the first bug of this kind we've encountered. But I can check.

amercader:
Thanks @mjanez , this all makes sense. 
Can you mention the namespace change in the [changelog](https://github.com/ckan/ckanext-dcat/blob/3afc1e2edd59efaced33980ccabd0a7d86e96d77/CHANGELOG.md?plain=1#L3). There's a small change that users with custom profiles using the old variant might have to update their code.

mjanez:
> Thanks @mjanez , this all makes sense. Can you mention the namespace change in the [changelog](https://github.com/ckan/ckanext-dcat/blob/3afc1e2edd59efaced33980ccabd0a7d86e96d77/CHANGELOG.md?plain=1#L3). There's a small change that users with custom profiles using the old variant might have to update their code.

Checks if it is sufficient: 4a47c101d17cfb48ba2c1d4230ca7b12141ad853

amercader:
Thanks @mjanez

PR #355, opened 2025-08-21
Improve access check in dcat_dataset_show

thorge:
Currently, the `dcat_dataset_show` action only calls:

```python
toolkit.check_access('dcat_dataset_show', context, data_dict)
```

However, the `dcat_dataset_show` auth function always evaluates to `True`, which is generally correct for DCAT endpoints. The problem arises when requesting a private dataset in a serialized format (e.g., `.ttl`).

Since no access check is performed for `package_show`, the call to:

```python
dataset_dict = toolkit.get_action('package_show')(context, data_dict)
```

can raise an uncaught auth exception. This results in a **500 Internal Server Error**.

**Proposed Fix:**
Add a `package_show` access check before calling the action:

```python
toolkit.check_access('package_show', context, data_dict)
```

**Steps to Reproduce:**

1. Create a private dataset.
2. Try to access its Turtle representation (e.g., `dataset/{id}.ttl`).
3. Observe that a 500 error is raised.

### DIFF
--- a/ckanext/dcat/logic.py
+++ b/ckanext/dcat/logic.py
@@ -20,6 +20,8 @@ def dcat_dataset_show(context, data_dict):
 
     toolkit.check_access('dcat_dataset_show', context, data_dict)
 
+    toolkit.check_access('package_show', context, data_dict)
+
     dataset_dict = toolkit.get_action('package_show')(context, data_dict)
 
     serializer = RDFSerializer(profiles=data_dict.get('profiles'))


thorge:
@amercader Yes, exactly, the `package_show` action does perform its own access check, but in `dcat_dataset_show` the exception is not caught, so it propagates all the way up to Flask, resulting in a 500 error:

```
Traceback (most recent call last):
  File "/usr/lib/ckan/default/lib/python3.8/site-packages/flask/app.py", line 1516, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/lib/ckan/default/lib/python3.8/site-packages/flask/app.py", line 1502, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**req.view_args)
  File "/usr/lib/ckan/default/src/ckanext-dcat/ckanext/dcat/blueprints.py", line 23, in read_dataset
    return utils.read_dataset_page(_id, _format)
  File "/usr/lib/ckan/default/src/ckanext-dcat/ckanext/dcat/utils.py", line 410, in read_dataset_page
    response = toolkit.get_action('dcat_dataset_show')({}, {'id': _id,
  File "/usr/lib/ckan/default/src/ckan/ckan/logic/__init__.py", line 580, in wrapped
    result = _action(context, data_dict, **kw)
  File "/usr/lib/ckan/default/src/ckanext-dcat/ckanext/dcat/logic.py", line 22, in dcat_dataset_show
    toolkit.check_access('package_show', context, data_dict)
  File "/usr/lib/ckan/default/src/ckan/ckan/logic/__init__.py", line 391, in check_access
    raise NotAuthorized(msg)
```

amercader:
@thorge Sorry, I'm not able to reproduce this. `NotAuthorized` exceptions are not catch by actions like `dcat_dataset_show`, but by higher level ones like the function that is called by the blueprint:

https://github.com/ckan/ckanext-dcat/blob/193bea79f73a8086c83aafab4e348e7b7c3572d5/ckanext/dcat/utils.py#L407-L411

`NotAhuthorized` should be catch here and the `abort()` call should gracefully return a 403 error. That's what I'm getting when requesting http://localhost:5000/dataset/test-private-dataset.ttl

Is this the full stacktrace? Have you modified the dcat endpoints or auth functions in some way?

thorge:
@amercader Thanks for taking a closer look! I’m currently working on an older branch for legacy Python support, so I missed some recent commits in the utils. That’s why the exception wasn’t caught. I was able to resolve the issue now. Sorry for taking up your time!

amercader:
@thorge no worries! Thanks for using the extension and reporting the issue

PR #365, opened 2025-10-24
Update base.py not only utf-8

piersoft:
There are some endpoint with special char. UTF-8 is not sufficient and harvesting not works

### DIFF
--- a/ckanext/dcat/harvesters/base.py
+++ b/ckanext/dcat/harvesters/base.py
@@ -96,7 +96,13 @@ class DCATHarvester(HarvesterBase):
                                             harvest_job)
                     return None, None
 
-            content = content.decode('utf-8')
+            # utf-8 it's not only codific. there are endpoint with special chars
+            if isinstance(content, bytes):
+                 try:
+                    content = content.decode('utf-8')
+                 except UnicodeDecodeError:
+                    content = content.decode('windows-1252', errors='ignore')
+
 
             if content_type is None and r.headers.get('content-type'):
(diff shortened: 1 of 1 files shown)


amercader:
@piersoft This makes sense, but `latin-1` is probably a safer fall back. And in fact we should check the response headers to see if there is a charset defined. What about something like:

```diff

diff --git a/ckanext/dcat/harvesters/base.py b/ckanext/dcat/harvesters/base.py
index b4affb4..4c25a70 100644
--- a/ckanext/dcat/harvesters/base.py
+++ b/ckanext/dcat/harvesters/base.py
@@ -1,5 +1,6 @@
 import os
 import logging
+import re

 import requests
 import rdflib
@@ -96,10 +97,21 @@ class DCATHarvester(HarvesterBase):
                                             harvest_job)
                     return None, None

-            content = content.decode('utf-8')
-
-            if content_type is None and r.headers.get('content-type'):
-                content_type = r.headers.get('content-type').split(";", 1)[0]
+            # Try charset from Content-Type header
+            encoding = 'utf-8'
+            if content_type_header := r.headers.get('content-type'):
+                match = re.search(r'charset=([^;]+)', content_type_header)
+                if match:
+                    encoding = match.group(1).strip().strip('"\'')
+
+            try:
+                content = content.decode(encoding)
+            except (UnicodeDecodeError, LookupError):
+                # Fall back to a safer option
+                content = content.decode('latin-1')
+
+            if content_type is None and r.content_type_header:
+                content_type = content_type_header.split(";", 1)[0]

             return content, content_type
```

PR #368, opened 2025-11-21
Update euro_dcat_ap_base.py

piersoft:
in my opinion the default is Attribution


amercader:
@piersoft Can you provide a description of the changes and the reasoning behind them? it helps when reviewing without any previous context

piersoft:
In EU, the default Licenses for Public Data are CC-BY or CC0 (dir. 1024/2019). So when a License is not provided, the fall back value it should be Attribution. No?

PR #371, opened 2026-01-24
Update processors.py add dcat:service in dcat:Catalog

piersoft:
dcat:service must be in dcat:Catalog if distribution have dcat:accessService


piersoft:
🐞 Missing dcat:service triples on dcat:Catalog (DCAT-AP 2.x / 3.x)

Describe the bug / enhancement

According to DCAT-AP 2.x and 3.x specifications, when a dcat:DataService is used (typically referenced from a dcat:Distribution via dcat:accessService), the corresponding dcat:Catalog should also explicitly list the service using dcat:service.

At the moment, ckanext-dcat correctly generates dcat:DataService nodes and links them from distributions via dcat:accessService, but never exposes them at catalog level using dcat:service.

This results in incomplete DCAT-AP catalogs and validation warnings/errors in EU and national validators (e.g. DCAT-AP, DCAT-AP_IT, EU MQA).

🔍 Current behavior

Example (simplified):

<distribution-1> dcat:accessService <service-1> .
<service-1> a dcat:DataService .

But no triple like:

<catalog> dcat:service <service-1> .


is ever generated.

✅ Expected behavior (DCAT-AP compliant)

The catalog should explicitly reference all exposed services:

<catalog> dcat:service <service-1> .


This is required by DCAT-AP 2.x / 3.x to make services discoverable at catalog level.

🧠 Root cause

In ckanext-dcat:

dcat:Catalog is serialized before or independently of datasets/distributions

dcat:DataService nodes are created later while serializing distributions

There is no post-processing step that connects generated DataService nodes back to the catalog

As a result, dcat:service is never added.

💡 Suggested solution

Add a step during catalog serialization (e.g. in serialize_catalog() in processors.py) that:

determines the actual target catalog for each dataset (main catalog or source catalog)

for each dataset, inspects its distributions

for each dcat:accessService, adds:

(catalog_ref, DCAT.service, data_service_ref)


This can be done safely after graph_from_dataset() and _add_source_catalog() are executed.

piersoft:
there is also a mini patch for access_rights in extra fields

PR #387, opened 2026-09-14
Fix dataset indexing failing when groups-field has repeating_subfields set

bzar:
If dataset schema sets `repeating_subfields` on `groups` field the iteration here fails because `groups` [is hardcoded to a list of strings in CKAN](https://github.com/ckan/ckan/blob/0731a7a4ab76dd11889eb89a892b3feae4fd69a8/ckan/lib/search/index.py#L176).

The fix here is not to enforce the schema implicitly and only provide best effort indexing by not processing fields in unexpected formats.


bzar:
Actually, this breaks groups indexing because later the field is popped from `dataset_dict`. We need to skip processing non-`list[dict]` fields altogether.

